PR #1, opened 2017-11-07
Added the option to plot TADs as triangles from a bed file

fidelram:


bgruening:
@fidelram thanks! Is this from the the py3k branch?

fidelram:
actually not! :(

PR #9, opened 2018-03-13
Fix for coolers

Phlya:
Fixed argument when using coolers


fidelram:
@Phlya Thanks for the PR. Did you succeeded using a cooler matrix with this change? I am under the impression that further changes are needed to properly implement this.

Phlya:
No, the issue I reported in HiCExplorer prevent me from using coolers, unfortunately.

PR #10, opened 2018-03-13
Update requirements.txt

joachimwolff:
This line needs at least HicExplorer version 2.0 (and in 2.1 it is failing because I remove the line which I shouldn't. Maybe version 2.1.1 should be the dependecy).
https://github.com/deeptools/pyGenomeTracks/blob/master/pygenometracks/tracksClass.py#L1726


bgruening:
@joachimwolff yes please release 2.1.1 and then we fix the conda packages.

PR #12, opened 2018-03-14
added citation

vivekbhr:
just adding citation for now, but I would keep the branch for possible RTD integration.


fidelram:
Vivek, can you add the citation text to the help message of the command line tools as well?

vivekbhr:
@fidelram done.

vivekbhr:
@fidelram python 2.7 error, shall I add # coding=utf-8 on top?

fidelram:
I really don't know how to solve it. Try to add the encoding in the first line. If it does not work, simply remove the tilde from my last name.

PR #16, opened 2018-03-23
added support for summary methods min and max for bigwig.

fidelram:
Also added a bigwig sample with missing data and fixed "nans to zeros" which was not working


pygenometracks/tests/test_bigWigTrack.py
```diff
@@ -0,0 +1,99 @@
+# -*- coding: utf-8 -*-
+import matplotlib as mpl
+mpl.use('agg')
+from matplotlib.testing.compare import compare_images
+from tempfile import NamedTemporaryFile
+import os.path
+import pygenometracks.plotTracks
+
+ROOT = os.path.dirname(os.path.abspath(__file__)) + "/test_data/"
```
bgruening: os.path.join()

pygenometracks/tests/test_data/bigwig.ini
```diff
@@ -0,0 +1,70 @@
+
+[test bigwig lines]
+file = bigwig2_X_2.5e6_3.5e6.bw
```
bgruening: this file does not need to be included, it is generated automatically, isn't it?

fidelram: no, is not generated automatically. They are part of the test_data

PR #18, opened 2018-03-26
Tabix support

fidelram:
Added tabix support as part of the bedGraphTrack. All tracks that inherit this track have access to either bedgraph like and tabix files. Furthermore, I added new features to the bedgraph format and improved the bedgraph_matrix format.


bgruening:
Its late Fidel, go to bed :)

fidelram:
I don't know what the coffee that I took had. This is so unusual!

PR #19, opened 2018-03-26
Dynamic loading of tracks

fidelram:
This PR changes the way tracks are loaded. Now, simply added a file with a track class to the pygenome/tracks folder is needed. Before, the class was added to the `trackClass.py` file.

I expect the new method to be simpler and easier to maintain.


fidelram:
The travis error for python3 is caused by a problem in HiCExplorer that is already been fixed.

Issue #32, opened 2018-06-21
difference between bigwig tracks


Phlya:
Hi,

Would be really cool if it was possible to subtract one track from another and show difference between them. Or log ratio. I suppose, ideally any mathematical operations :) Thanks!

Ilya

dpryan79:
See `bigwigCompare` from deepTools and `wiggletools` as ways to preprocess your data, though I realize doing this sort of thing on the fly would be nice.

Phlya:
Yeah, thanks, I have found it actually, but yes, on the fly would be very convenient.

Phlya:
And a related to that question is would it be possible to have a different colour for negative values? I think currently there is no such option, looking at the code?

fidelram:
The way to have different colors is to set two tracks with no space in
between. For the upper track you set the scale from 0 to 'auto' or any
other value, for the lower track you set the scale from 'auto' to 0. Each
track can have it's own color.

Phlya:
Thanks! Yeah, that's possible, but seems a little hacky to me, and it's especially complicated since I am using pygenometracks as a library and would also need to set up the axes manually... I think I could quite easily add the negative colour option to pygenometracks - if I do, would you consider a PR for this?

fidelram:
Sure, If you can do such a thing I will be happy to merge.

lldelisle:
After nearly 2 years, this is in version 3.4

PR #33, opened 2018-06-22
Different color for negative values

Phlya:
Hi, as per https://github.com/deeptools/pyGenomeTracks/issues/32 - different colour for negative values. I started with just bigwig plots to see whether you agree with the strategy, and if it looks OK to you I can apply the same changes to bedgraph tracks too.

### DIFF
--- a/pygenometracks/tracks/BedGraphTrack.py
+++ b/pygenometracks/tracks/BedGraphTrack.py
@@ -47,6 +47,9 @@ file_type = {}
         if 'color' not in self.properties:
             self.properties['color'] = DEFAULT_BEDGRAPH_COLOR
 
+        if 'negative color' not in self.properties:
+            self.properties['negative color'] = self.properties['color']
+
         if 'nans to zeros' not in self.properties:
             self.properties['nans to zeros'] = False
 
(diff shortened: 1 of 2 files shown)


fidelram:
Thanks for adding this, I will be happy to merge! And sorry for the delay I am travelling.

Can you fix the travis error? it is caused by some missing spaces around operators. 

Also, can you perhaps add an image to show the results?

Phlya:
Thanks for the feedback, I will do that. BTW currently I have got a problem that my developmental installation doesn't work anymore due to this https://github.com/deeptools/HiCExplorer/issues/261

Phlya:
To add an image, should I add some example data to the repository too? I'm presuming, you mean an image for readme?

fidelram:
Thanks for fixing the errors. I was only thinking to you could add an image here, such that I can get an impression on how this is working. But if you want to edit the readme file and add and image that would be great too.

Phlya:
Right. OK, yeah, I'll do that.

Phlya:
Adding it to the readme is a bit complicated because I don't think there is a good example date file there already, so I would have to add some data, and I work with mouse, not Drosophila, for which all the current examples are, so it would look a little strange, I think.

Phlya:
image
Here is a simple example with bigwigs (loss of K27me3 across HoxC in ES cells in 2i culture).

Phlya:
There is a problem with `type = line` though, zero crossing are not shown and this makes the plot look weird - and fixing this is not trivial, would require interpolation to find positions of zero crossings and then extending the lines to there.
image
I tried doing this using fill_between with transparent facecolor, but then it show the line at zero with the same color, which is a little weird (just a small test example without pyGenomeTracks to show the problem). 
image
An option is to then add a line to show the zero level, which might be desirable anyway:
image
For now I added a warning that line plots with a different negative colour might produce ugly plots, but let me know if you would prefer a different solution.

fidelram:
Thanks for the images. For the `type=line` I assume that when no negative color is given the result is as previous right? I am asking because in the code, when no negative color is given, then the same positive color is asigned. In the case of type=line, this could  cause the line to break as in the in the example that you sent. 

can you check this?

Phlya:
See the commit above.

PR #37, opened 2018-07-08
added test for negative color from bigwig or bedgraph tracks

fidelram:


fidelram:
@Phlya I added some tests for the new feature that you added. However, I see some problems with the bigwig fill case. See the first track in this image:

image

Can you take a look? I added a bigwig file that has some negative values to the test data. You can your the test as `py.test test_bigWigTrack.py`. 

In the bedgraph case I don't see the problem, but for type=line and points the image looks strange to me:

image

Phlya:
Thank you, I'll try to have a look tomorrow!

Phlya:
Sorry, I didn't manage to look into it then, and now I'm on holidays with not much free time. But I remember about this.

Phlya:
OK, for some reason `interpolate=True` creates this weird problem for bigwig tracks... Strange. I'll try some simple examples with matplotlib plots, maybe it's a bug...

Phlya:
image

Yep, pretty sure this is a bug in matplotlib!

fidelram:
I think that setting the values to zero may work.

fidelram:
I would like to add a new pyGenomeTracks release. Do you think we can finalize this? Using `interpolation=False` does not work?

Phlya:
Hi, sorry, forgot about this to be honest. See my PR, nans get replaced by zeros. I ran the test and it looks fine.

Phlya:
FYI, as discovered in mpl issue tracker, updating numpy to 1.16 fixes the issue with interpolation.
https://github.com/matplotlib/matplotlib/issues/11781#issuecomment-469375173

fidelram:
Great to know. I will update the requirements

Phlya:
Well, we circumvented the issue by converting all NaNs to zeros, so it fine now anyway, but you might want to revert https://github.com/deeptools/pyGenomeTracks/pull/37/files/47b13185690f65e53f27f92666095ed3ab5d2048..HEAD then if you are updating requirements.

Issue #31, opened 2018-06-20
color_only_init error


Phlya:
When using pygenometracks to plot Hi-C data (at least from cooler files) I get an error:
```
INFO:root:Generating grammar tables from /usr/lib/python3.5/lib2to3/Grammar.txt
INFO:root:Generating grammar tables from /usr/lib/python3.5/lib2to3/PatternGrammar.txt

title not set for 'section 3. [Genes]'
Traceback (most recent call last):
  File "/usr/local/bin/pyGenomeTracks", line 11, in <module>
    main(args)
  File "/usr/local/lib/python3.5/dist-packages/pygenometracks/plotTracks.py", line 279, in main
    track_label_width=args.trackLabelFraction)
  File "/usr/local/lib/python3.5/dist-packages/pygenometracks/tracksClass.py", line 102, in __init__
    self.track_obj_list.append(track_class(properties))
  File "/usr/local/lib/python3.5/dist-packages/pygenometracks/tracks/HiCMatrixTrack.py", line 50, in __init__
    self.hic_ma = HiCMatrix.hiCMatrix(self.properties['file'], color_only_init=True)
TypeError: __init__() got an unexpected keyword argument 'color_only_init'

```

fidelram:
It should be `cooler_init_only` 
@joachimwolff can you take a look at this? Do you think we can add a test for cooler?

joachimwolff:
Again this error? I thought it is fixed with 2.1 version. I will have a look at it.

bgruening:
@Phlya  please give the latest version a try. It should hopefully solve your issue.
You can try the newest version here: https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/iuc/pygenometracks/pygenomeTracks/

PR #41, opened 2018-09-04
Replacing dependency to HiCExplorer by HiCMatrix

joachimwolff:
- Replace dependency of HiCExplorer by HiCMatrix.
- Fixing cooler read support, see #31

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,8 +48,8 @@ before_install:
 
 # Install packages
 install:
-  - conda install --yes -c conda-forge -c bioconda python=$TRAVIS_PYTHON_VERSION --file requirements.txt
-  - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then conda install --yes -c defaults -c conda-forge -c bioconda configparser ; fi
+  - conda install --yes -c bioconda -c conda-forge python=$TRAVIS_PYTHON_VERSION --file requirements.txt
+  - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then conda install --yes -c bioconda -c conda-forge -c defaults configparser ; fi
   - python setup.py install
 
 # command to run tests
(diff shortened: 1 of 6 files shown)


fidelram:
I would like to prepare a new release. Do you think we should include these changes?

joachimwolff:
It would be great if we could include them. It makes it possible to replace the content of hicPlotTADs with a call to pyGenomeTracks and therefore brings all new features and bugfixes to HiCExplorer.  With the current version we would have a mutual dependency of pyGenomeTracks and HiCExplorer.

fidelram:
Thanks!

joachimwolff:
@dpryan79 Do you have any idea why the bigwig test case with macOS is failing?

dpryan79:
For some reason installing `configparser` results in numpy getting downgraded to an incompatible version for some bioconda packages. You can see that here: https://travis-ci.org/deeptools/pyGenomeTracks/jobs/426412361#L525-L542

Now you're probably asking why installing configparser would affect numpy/scipy/etc. I have no clue about that, it makes no sense. The fix would then be to install conigparser first and then the rest of the packages.

BTW, you could also stop testing python 2.7, it's going to die soon and numpy et al. are dropping support in a few months.

fidelram:
When I have run into compatibility issues like this one, the way to go is to put all required modules in one file and conda takes care to install the common requirements. So, we can have a file called requirements2.7.txt that includes configparser and see if this solves the problem.

python 2.7 will be retired in 2020. I suggest we at least make this release compatible with python 2.7.

joachimwolff:
My own, probably not reproducible, tests:
- Creating a new environment with python 2.7.15 and configparser --> No dependency to numpy
- Installing in an environment with already installed numpy --> Downgrade to numpy 1.13, sometimes 1.11
- Installing configparser in an environment without already installed numpy but installing numpy at the same time with newest version --> Dependency but it's ok install the newest version.

I do not understand this dependency cycle. Anyone? 
However, the last test case passes now by enforcing the newest numpy version in the requirements.txt file.

dpryan79:
Conda's dependency resolution is truly a mystery.

PR #42, opened 2018-09-05
Fix interpolation in BigWig fill_between

Phlya:
This should fix the problem until matplotlib gets fixed


Phlya:
@fidelram Hi, as far a I can tell this fix was never merged into master, while the neg colour option has been, so currently in the released and master version the plots that contain neg values have this problem.

Phlya:
Hi. This somehow didn't make it into master, it seems...

PR #46, opened 2018-10-24
Fixing bug concerning log and -log option for Hi-C matrices

joachimwolff:
It is `mask == False` and not `mask is False`. First one checks if the entries of `mask` are `False` and creates a boolean array with the result , second one checks if `mask` object is `False`. Adding catch for possible case all values in the matrix are 0.

### DIFF
--- a/pygenometracks/tracks/HiCMatrixTrack.py
+++ b/pygenometracks/tracks/HiCMatrixTrack.py
@@ -196,13 +196,19 @@ file_type = {}
 
             elif self.properties['transform'] == '-log':
                 mask = matrix == 0
-                matrix[mask] = matrix[mask is False].min()
-                matrix = -1 * np.log(matrix)
+                try:
+                    matrix[mask] = matrix[mask == False].min()
+                    matrix = -1 * np.log(matrix)
+                except Exception:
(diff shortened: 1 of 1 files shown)


pygenometracks/tracks/HiCMatrixTrack.py
```diff
@@ -196,13 +196,19 @@ def plot(self, ax, chrom_region, region_start, region_end):
 
             elif self.properties['transform'] == '-log':
                 mask = matrix == 0
-                matrix[mask] = matrix[mask is False].min()
-                matrix = -1 * np.log(matrix)
+                try:
+                    matrix[mask] = matrix[mask == False].min()
+                    matrix = -1 * np.log(matrix)
+                except Exception:
```
bgruening: @joachimwolff should we catch a more specific exception here?

fidelram: Are you sure that the exception only happens if all values are cero? In that case you can simply check for `np.all(matrix == 0) `

joachimwolff: Can also be the case if the matrix contains no data.

PR #68, opened 2019-05-02
update import

shouldsee:


bgruening:
I think this should be `from pygenometracks.tracks import GenomeTrack` in develop. I changed that. Thanks for your contribution.

PR #69, opened 2019-05-02
update class name

shouldsee:


bgruening:
Thanks @shouldsee! I fixed that in develop. I hope that is ok! We will release a new version soon.

PR #82, opened 2019-07-19
hacky xaxis overlap fix

djakubosky:
Here is a hacky fix to stop the xaxis ticklabels from overlapping the chromosome label.  I think a better solution would be to allow the user to increase the height of the Xaxis track such that the text could comfortably fit with your original settings, but I couldn't figure out where such a thing was set.  In principle, this fix will cause text to extend below the bounds of a default XaxisTrack, thus inserting a spacer below the track is necessary too.


joachimwolff:
Hi,

To accept a PR all test cases and linting needs to pass, please take care of this.

Thanks a lot,

Joachim

bgruening:
@djakubosky could you please rebase this against latest develop?

PR #83, opened 2019-07-23
Add gtf possibility and other small things

lldelisle:
1) Use gtf:
a) I wrote a new class ReadGtf to make an iterator from a gtf file.
b) In BedTrack.py, I added the gtf extension and use ReadGtf if the extension match or if type='gtf' and use ReadBed if not.
c) Added to the requirement gffutils but I do not know if it is useful...
2) Small correction:
I noticed that for bed12 when the thick start = thick end = chrom_start the first exon is plotted as coding and if  start = thick end = chrom_end the last exon is plotted as coding. So I added a condition in both draw_gene_with_introns_flybase_style and draw_gene_with_introns.

What I can do:
add an example,
tell me if you need something else to merge the PR...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,6 @@ htmlcov
 .mr.developer.cfg
 .project
 .pydevproject
+
+# vscode
+.vscode
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ pyGenomeTracks can make plots with or without Hi-C data. The following is an exa
 
 Installation
 ------------
-pyGenomeTracks works with python 2.7 and python 3.6.
+pyGenomeTracks works with python >=3.6.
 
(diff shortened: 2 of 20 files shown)


joachimwolff:
Hi Lucille,

To accept this PR all test cases and linting needs to pass, please take care of this. Moreover, for all new added features there need to be test cases in pygenometracks/tests.

Thanks a lot,

Joachim

bgruening:
@lldelisle you need to rebase it against develop. There are a few commits included that do not belong here I guess.

Adding gffutils as requirement makes a lot of sense imho, if we start to dupport GFF files. Thanks for working on it.

lldelisle:
I was afraid you did not want to use more bed formats so I separated by PR but in fact if you accept #84 there is no reason to work on this one.

PR #84, opened 2019-07-23
Add other bed

lldelisle:
I changed readBed.py to have the possibility to plot bed4 (only names), bed5 (names and score), these will be converted in BedInterval with 6 fields added the default values ".", 0, "." as it was the case for bed3.
I also modified the code to be able to put bed8 and non-standard bed reduced to bed6 without having an error.

In BedTrack.py, I only changed the SUPPORTED_ENDINGS. 
The rest is just formatting to help me to read and fit the flake8 line length.

What I noticed, it that when the bed_type is bed9, the thickstart and thickend are not used. Shouldn't we convert everything which is above bed6 (bed8 and bed9) to bed12 and use the draw_gene_with_introns_flybase_style or draw_gene_with_introns functions?

Also, I have a question: where "type=genes" is used?

Many thanks,
I really like this project.


lldelisle:
(I made it on top of the gtf but this is independent)

LeilyR:
please change your base to develop, thanks!

lldelisle:
You mean, I should PR myself to the master branch before doing the PR?

lldelisle:
Or to the develop branch?
Sorry I got it.

lldelisle:
Oups I think I did the wrong way...

LeilyR:
sorry for not being clear I have meant sending your PR to the develop branch

joachimwolff:
Hi Lucille,

To accept this PR all test cases and linting needs to pass, please take care of this. Moreover, for all new added features there need to be test cases in pygenometracks/tests.

Thanks a lot,

Joachim

lldelisle:
I needed to solve the issue #85 because travis was failing (this is commit [01c17cb](https://github.com/deeptools/pyGenomeTracks/pull/84/commits/01c17cbd6be0044e5e80482ebf6b08dec0ec2319)).
I added a test with a gtf and a bed4.

> What I noticed, it that when the bed_type is bed9, the thickstart and thickend are not used. Shouldn't we convert everything which is above bed6 (bed8 and bed9) to bed12 and use the draw_gene_with_introns_flybase_style or draw_gene_with_introns functions?
> 
> Also, I have a question: where "type=genes" is used?

Thanks,

Lucille

lldelisle:
Because I also added the gtf, here were my comment on the gtf part:
1) Use gtf:
a) I wrote a new class ReadGtf to make an iterator from a gtf file.
b) In BedTrack.py, I added the gtf extension and use ReadGtf if the extension match or if type='gtf' and use ReadBed if not.
c) Added to the requirement gffutils...
2) Small correction:
I noticed that for bed12 when the thick start = thick end = chrom_start the first exon is plotted as coding and if  start = thick end = chrom_end the last exon is plotted as coding. So I added a condition in both draw_gene_with_introns_flybase_style and draw_gene_with_introns.

lldelisle:
I corrected the interval height in the bedtrack but I think it is not used because the height is forced. I think if we want to use this, we need to update get_tracks_height in tracksClass.py but this would be very difficult.

setup.py
```diff
@@ -99,7 +99,8 @@ def checkProgramIsInstalled(self, program, args, where_to_download,
                        "pyBigWig >=0.3.7",
                        "hicexplorer >= 2.1.1",
                        "pysam>=0.14",
-                       "pytest"
+                       "pytest",
+                       "gffutils"  # I do not know which version
```
joachimwolff: 0.9 is the current one.

Issue #85, opened 2019-07-23
Travis fail with matplotlib3.1.0


lldelisle:
Hi,
I noticed that travis failed for this line with matplotlib3.1.0:
https://github.com/deeptools/pyGenomeTracks/blob/176455b3503be474215f9a930ad090bde84368cc/pygenometracks/tracks/BedGraphMatrixTrack.py#L95
In fact, this line was not executed in previous versions of matplotlib (it was ignored because the second parameter was not correct (which should be major, minor or both):
https://github.com/matplotlib/matplotlib/blob/93990abc5c6917a6a82d966ce1a4380d30370b00/lib/matplotlib/axes/_base.py#L2726-L2757
I guess what you wanted to but was: 
ax.grid(True, axis='y')
but I am not sure.

Issue #79, opened 2019-06-20
Missing y-axis scale on hicPlotTADs


StephenRicher:
Hi,

I am making great use of this tool but am having trouble with plotting a scale bar on the HiC matrix generated via hicPlotTADs.

My figure looks the same as one of your test images at: https://github.com/deeptools/HiCExplorer/blob/master/hicexplorer/test/test_data/hicPlotTADs/pygenometracks.png 

The colour scale bar is present but no numbers are there. I can't find the .ini file associated with the above test data plot but I was wondering if there was a setting in the configuration file that is causing these to not be plotted? As in the above figure all my other tracks plot the scales correctly.

Thanks,
Stephen

vivekbhr:
@StephenRicher are you able to reproduce this issue with pyGenomeTracks? If yes. please let us know the version number, config file and an example output.

StephenRicher:
Hi,

Yep I have the same issue - I've attached a basic config file and test output below. I am using pyGenomeTracks 3.0. 

Thanks,
Stephen

![test](https://user-images.githubusercontent.com/45041598/59927957-8e1db580-9435-11e9-8630-ccc69db39863.png)

[tracks.txt](https://github.com/deeptools/pyGenomeTracks/files/3314761/tracks.txt)

vivekbhr:
Yes this looks like a known issue. The labels were missing for ticks when you use log1p with certain matplotlib versions. I think updating your matplotlib would help

kimj50:
hi, I also ran into a similar problem. hicexplorer or pygenometrack doesn't show the tickmarks.
I am currently using: pygenometracks v3.0, matplotlib v3.0.3. 
track:
[x-axis]
where = top

[hic matrix]
file = /scratch/kimj50/elegans/matrix_normalize/N2_2017_emb_sum_5kb_correct_normalize.h5
title = test
depth = 1000000
transform = log

[test.pdf](https://github.com/deeptools/pyGenomeTracks/files/3378522/test.pdf)

lldelisle:
Hi, I also have the same issue with pyGenomeTracks 3.0 with matplotlib v3.1.1 but also v3.0.3, and v3.0.1...

lldelisle:
This is now solved since v3.1.

PR #88, opened 2019-08-02
Matplotlib3.1.1

lldelisle:
I solved the #85 and #79 linked to the matplotlib changes from 2 to 3.


pygenometracks/tracks/EpilogosTrack.py
```diff
@@ -73,8 +73,11 @@ def plot(self, ax, chrom_region, start_region, end_region):
         """
         from matplotlib import cm
         cmap = cm.get_cmap('tab20b')
+        try:
+            values_list, pos_list = self.get_scores(chrom_region, start_region, end_region)
+        except TypeError:
+            return
```
bgruening: Should we omit here a warning at least?

pygenometracks/tracks/HiCMatrixTrack.py
```diff
@@ -256,29 +262,33 @@ def plot_y_axis(self, cbar_ax, plot_ax):
             formatter = LogFormatter(10, labelOnlyBase=False)
             aa = np.array([1, 2, 5])
             tick_values = np.concatenate([aa * 10 ** x for x in range(10)])
-            cobar = plt.colorbar(self.img, ticks=tick_values, format=formatter, ax=cbar_ax, fraction=0.95)
+            try:
+                cobar = plt.colorbar(self.img, ticks=tick_values, format=formatter, ax=cbar_ax, fraction=0.95)
+            except AttributeError:
+                return
```
bgruening: Same here ... a warning?

pygenometracks/utilities.py
```diff
@@ -9,8 +10,9 @@ def to_string(s):
     if isinstance(s, str):
         return s
     if isinstance(s, bytes):
-        if sys.version_info[0] == 2:
-            return str(s)
+        assert(sys.version_info[0] != 2)
+#        if sys.version_info[0] == 2:
```
bgruening: can be removed

pygenometracks/utilities.py
```diff
@@ -21,8 +23,9 @@ def to_bytes(s):
     """
     Like toString, but for functions requiring bytes in python3
     """
-    if sys.version_info[0] == 2:
-        return s
+    assert(sys.version_info[0] != 2)
+#    if sys.version_info[0] == 2:
```
bgruening: remove?

setup.py
```diff
@@ -97,9 +97,8 @@ def checkProgramIsInstalled(self, program, args, where_to_download,
                        "matplotlib >= 2.0.0",
                        "intervaltree >= 2.1.0",
                        "pyBigWig >=0.3.7",
-                       "future >= 0.16.0",
-                       "hicmatrix",
-                       "pysam",
+                       "hicexplorer >= 2.1.1",
```
bgruening: please adopt also numpy to the one in requirements

LeilyR: I have removed as many as i could find of python2 stuffs while ago. I would have assumed it is already part of the current release.

lldelisle: Yes, this is LeilyR commit but it was because the develop branch was behind master by one comit...
@LeilyR, in fact, I do not see why we require hicexplorer >= 2.1.1. For me, the former were more close to what is actually needed, no?
@bgruening hicexplorer or hicmatrix require numpy so I do not know if we should specify here...

LeilyR: I am not sure if hicexplorer is needed at all. I think only hicmatrix should be there. I can check.

bgruening: What I mean is that `"numpy >= 1.12.1"` is different from `numpy >= 1.16`. So setup.py is different from requirements.py.

@joachimwolff can comment more on what is currently used for hicex.

lldelisle: @LeilyR, yeah, I mean I would put hicmatrix, maybe future... In the conda recipe there is no pygenometracks:
```
requirements:
  host:
    - python >=3.6
    - pip
  run:
    - python >=3.6
    - numpy >=1.16
    - matplotlib >=3.0
    - intervaltree >=2.1.0
    - pybigwig >=0.3.4
    - future >=0.17.0
    - hicmatrix >=9
    - pysam
```

lldelisle:
Hi,
Do you think one of you could merge master to develop so we could see better what is already done?
Many thanks,

bgruening:
I'm tempted to merge this and our other PR to develop and then continue from there with the aim of release something 3.0is next week.

@LeilyR @fidelram @joachimwolff any objections?

bgruening:
@lldelisle this now needs to small rebase.

lldelisle:
I am on holidays but feel free to do it.

bgruening:
GitHub is not updating this site accordingly. I think everything is now in develop, I merged, at least I tried everything into develop. Thanks!

I think time for a new release.
If anyone wants to test the develop branch intensively.

PR #89, opened 2019-08-11
3.1 release

bgruening:
* add the ability to control alpha of bedgraph tracks (thanks to @mikewolfe)
* add GTF supper (thanks to @lldelisle)
* add support for various BED flavors (thanks to @lldelisle)
* adopt codebase to lastest matplotlib changes (thanks to @lldelisle)
* various new tests, code cleanups and smaller fixes all over

Thanks a lot, everyone!


joachimwolff:
Python 2 support is gone since the last release, which was 3.0. I am a bit surprised these two things should be included in this PR again?

bgruening:
@joachimwolff changed that. It's wired this commit still is shown here.

setup.py
```diff
@@ -99,7 +99,8 @@ def checkProgramIsInstalled(self, program, args, where_to_download,
                        "pyBigWig >=0.3.7",
                        "hicexplorer >= 2.1.1",
```
joachimwolff: There is no dependency to this HiCExplorer version. Please change it to hicmatrix and check the other dependencies as well.

setup.py
```diff
@@ -99,7 +99,8 @@ def checkProgramIsInstalled(self, program, args, where_to_download,
...
                        "pysam>=0.14",
-                       "pytest"
+                       "pytest",
+                       "gffutils"  # I do not know which version
                        ]
 
 if sys.version_info[0] == 2 or (sys.version_info[0] == 3 and sys.version_info[1] == 4):
```
joachimwolff: Not required anymore.

bgruening: @joachimwolff feel free to commit directly to the branch.

pygenometracks/tracksClass.py
```diff
@@ -477,11 +477,12 @@ def plot(self, ax, chrom_region, region_start, region_end):
...
-            ax.axis["x"] = ax.new_floating_axis(0, 0.9)
-            label_y_pos = 0.01
-            vert_align = 'bottom'
+            ax.axis["x"] = ax.new_floating_axis(0, 0.99)  # create a little more room
+            label_y_pos = 0
+            vert_align = 'top'  # this probably causes text to go beyond axis lims desired- but makes more space
         ax.text(0.5, label_y_pos, chrom_region, horizontalalignment='center',
-                fontsize=int(self.properties['fontsize']), verticalalignment=vert_align, transform=ax.transAxes)
+                fontsize=int(self.properties['fontsize']), verticalalignment=vert_align, transform=ax.transAxes,
+                clip_on=False)  # in practice I need to set another spacer below to avoid clipping
```
lldelisle: For me this is strange to put the vertical alignment to top and then use a spacer to avoid clipping. Shouldn't we just increase the height of the xaxis to have enough space to avoid overlay?

bgruening: Sounds sensible. But matplotlib can yield to unexpected results :) Let's try it?

lldelisle: When I use this version without adding a spacer where there are few tracks like one or 2 the chromosome name is hidden or half hidden (see #93 the epilogos in examples)...

bgruening:
Ok, lets merge this and draft a release :)

lldelisle:
Many thanks, I think you can also delete the matplotlib branches...

PR #91, opened 2019-08-20
Unify dependencies

lldelisle:
Hi,
I modified the files to be consistent between the versions and the dependencies between setup.py, requirements.txt and environment.yml.


lldelisle:
- The only PR still open is the one with the x axis. I made a [PR](https://github.com/djakubosky/pyGenomeTracks/pull/1) to the author but he did not reply... Should I open a new PR to develop or should we wait?

- For the develop branch I suggest to regenerate all pictures (test and examples) with the develop version so people know what they can expect to have with the current version. But for this we should wait the last minute...

bgruening:
> The only PR still open is the one with the x axis. I made a PR to the author but he did not reply... Should I open a new PR to develop or should we wait?

New PR seems fine to me. The commits should stay so everyone is acknowledged and in theory, the PR should also close themselves if all commits are merged. Let's try it.

> For the develop branch I suggest to regenerate all pictures (test and examples) with the develop version so people know what they can expect to have with the current version. But for this we should wait the last minute...

I need to check how much work that is. Thanks for the suggestion!

lldelisle:
> > For the develop branch I suggest to regenerate all pictures (test and examples) with the develop version so people know what they can expect to have with the current version. But for this we should wait the last minute...
> 
> I need to check how much work that is. Thanks for the suggestion!
I think this should not be so hard if we decide to keep the ini files and generate a table with ini path, arguments and output... I can generate such a table and I have already the develop version installed locally.

bgruening:
> I think this should not be so hard if we decide to keep the ini files and generate a table with ini path, arguments and output... I can generate such a table and I have already the develop version installed locally.

Awesome, thanks a lot. I will try to set up the auto-deploy. So we need to just tag a release going forwards and everything else happens automatically.

PR #93, opened 2019-08-20
Update png

lldelisle:
Hi,
I added some ini files and a bash script to generate the png files. I also regenerated most of png.
Unfortunately I discovered that some plots have issues with the label of the chromosome being hidden because too low. I will try to find a way to make it appears well...


lldelisle:
@djakubosky,
I found a new way to solve the x-axis issue. If you want we can discuss it.
Lucille

lldelisle:
I think now it should be ok.

bgruening:
Merci!

PR #95, opened 2019-09-13
Narrow peaks

lldelisle:
Hi,

- I corrected a bug for use summit (use summit in the config file and use summits in the code)...
- I added a new test to show how use summit could indeed change the aspect
- I added a vertical bar in the narrowPeaks when plotted as box to be able to see the summits.


temp.yml
```diff
@@ -0,0 +1,14 @@
+name: pygenometracks_test
```
bgruening: is that one needed?

lldelisle: Oups, hopefully you spotted it.

bgruening:
Do you need a release?

lldelisle:
Well, if you don't mind, I would say yes, then I update the wrapper for galaxy.

lldelisle:
Many thanks

lldelisle:
I do not remember, it will autodeploy or only big releases?

bgruening:
Yes it should, but it was not ... I tried to fix it and found that I missed bumping the version number in the _version.py file. So you be there now. Can you remind an old man next time to bump the version number :)

PR #100, opened 2019-09-23
Update __init__.py

joachimwolff:
Setting several logging levels from dependencies from `INFO` to `WARNING`.


lldelisle:
@bgruening, the pip updated automatically but not bioconda, do you know why? Is it every night or something like this?

dpryan79:
The bot was down for a couple days, there may be a backlog of pull requests it needs to make.

lldelisle:
It is done... Thanks

Issue #26, opened 2018-05-29
gene label not shown when gene number to large


silvewheat:
I have two gene file, one file contains 59 genes, anther contains 89 genes. When I draw tracks, the first genes set can show gene label, the other can't. And, I have set the "labels = on" in both track.
It seems caused by the gene number.
I also try to plot with a single gene file contain more than 100 genes, the label can't shown either.
![image](https://user-images.githubusercontent.com/19344997/40636881-425245d6-6334-11e8-9f78-fdbd5b64f1b8.png)

vivekbhr:
The number of genes should not affect the plotting of gene names. Can you paste the `head` of both the files?

silvewheat:
These two files are all in bed12 format that I used [ea-utils/gtf2bed](https://github.com/ExpressionAnalysis/ea-utils/blob/master/clipper/gtf2bed) transform from gtf file.
The first file (59 genes, label showed):
![image](https://user-images.githubusercontent.com/19344997/40647241-f97f20c0-635d-11e8-8abe-1b979b8f89ab.png)
The second file (89 genes, label unshowed):
![image](https://user-images.githubusercontent.com/19344997/40647435-6f33e634-635e-11e8-9237-2fac3859fcc7.png)

fidelram:
I think that there is a limit of the number of genes with label that can be
seen. The threshold, if I remember correctly is 50. To view the labels
simple zoom in a bit.

silvewheat:
I tried it, but it seems really no label.
And I also change the font size in .ini file, also didn't work.

vivekbhr:
@silvewheat you might have to share the files with us to reproduce this issue. Do you mind attaching the files here?

silvewheat:
Hello,
   I hvae upload my .ini, .bed ,and the result .bed file in the package.
  My command line is:
    pyGenomeTracks --tracks tracks.ini --region chr1:100732629-102527660 --outFileName genes.pdf
[debug.zip](https://github.com/deeptools/pyGenomeTracks/files/2086615/debug.zip)

Thanks!

shenscore:
Hi, I met the same problem.
Have you found any resolutions ?

Thanks!

silvewheat:
I reduced the number of genes.
╮(╯﹏╰）╭

LeilyR:
[...]

PR #110, opened 2019-10-23
Nb of genes

lldelisle:
- add the number of genes above which label is switched to off as a parameter (max_labels) relative to #26.


pygenometracks/tests/test_bed_and_gtf_tracks.py
```diff
@@ -98,3 +128,16 @@ def test_plot_tracks_bed_and_gtf():
...
     os.remove(outfile.name)
+
+
+def test_plot_tracks_bed_with_maxLab():
+
+    outfile = NamedTemporaryFile(suffix='.png', prefix='pyGenomeTracks_test_', delete=False)
+    args = "--tracks {0}/bed_maxLab_tracks.ini --region X:2000000-3500000 --trackLabelFraction 0.2 --width 38 " \
+           "--dpi 130 --outFileName  {1}".format(ROOT, outfile.name).split()
+    pygenometracks.plotTracks.main(args)
+    print("saving test to {}".format(outfile.name))
```
bgruening: this can be removed I think

lldelisle: You mean you would not do a test for this functionality so I remove the test, the ini file and the png?

LeilyR: I wouldn't remove any test.  Let's see what Bjoern thinks, maybe he found it redundant with another test?

bgruening: Oh no tests are great, I love tests, but the print is not needed right?

PR #111, opened 2019-10-24
Use alpha more broadly

lldelisle:


pygenometracks/tests/test_data/alpha.ini
```diff
@@ -46,5 +50,36 @@ type = fill
...
+[spacer]
+
+[test bigwig]
+file = bigwig2_X_2.5e6_3.5e6.bw
+height = 7
+title = (bigwig color=red alpha = 0.5 max) overlayed with (bigwig mean color=green alpha=0.5 line:2) overlayed with (bigwig min color=blue alpha=0.5 points:2)
+color = red
+color = red
+alpha = 0.5
+summary method = max
```
bgruening: @lldelisle @LeilyR before we hit version 1.0 should we replace all those "two works" attributes with `*_*`? I kind of scares me always when I see it ... but this might be just me.

LeilyR: I totally agree i would replace i 2 words att with word_word. I will work on it.

lldelisle: I am going to propose a PR where
the value of the 'boolean' parameters can be everywhere, on, off, yes, no, 0, 1, true, false with no case sensitivity.
In this one I could also put that all spaces are replaced by `_` in `self.properties.keys` so that summary method will work but also summary_method. What do you think?

bgruening: I like the idea of supporting both, but I think we should deprecate the old style and remove it at some point.

> on, off, yes, no, 0, 1, true, false

It's probably more useable, but also more confusing for users that see other ini files? I'm +-0 on this. What would be nice is to have some kind of schema and a linting of our config files, so people can check without executing.
I just realize this should be an issue and not a random comment :)

lldelisle: I propose to move the discussion to an issue as this is not directly linked to this PR...

PR #113, opened 2019-10-24
Hic mat

lldelisle:
- You can choose to not rasterize the matrices of bedgraphmatrix and hicmatrix tracks.
- You can use a personnal colormap using an array with the different colors.
- The HiC matrix is now really 45 degrees when no height is given.
- The height put in the config file are now really respected (the total figure height is slightly bigger than the sum of all tracks to take into account the margin).
- fixed a bug where the ticks of the color scale bar were not at the good position due to margin changes at the end.


LeilyR:
Do you only have test for the colormap? what about rasterizing and height vs. no-height cases?

lldelisle:
For rasterizing, I can do a test, the output needs to be svg or pdf, I do not know how to check pdf/svg files are the same. Could you help me with that?
For the 45degrees, it is more a small correction. It was already close but it was 48 degrees instead of 45 degrees because the figure height was the sum of all height instead of being the sum of all height + the margin sizes. So I did not think a test was needed but I can do one.

lldelisle:
The example for rasterize and height has been added.
I pulled the develop branch.

pygenometracks/tests/test_data/browser_tracks_hic.ini
```diff
@@ -83,8 +83,10 @@ height = 0.5
 
 [hic matrix]
 file = Li_et_al_2015.h5
-title = depth=200000; show_masked_bins=yes;
+title = depth=200000; show_masked_bins=yes; colormap = ['blue', 'yellow', 'red']; max_value = 150
```
bgruening: the colormap should be remove here?

bgruening: Ah I see ...

LeilyR:
@lldelisle Is it possible that you have forgotten to `git add browser_tracks_cool.ini`. I cannot find it. I don't get how tests have been passed.

lldelisle:
Hi,
This file is created in the first part of the test:
https://github.com/deeptools/pyGenomeTracks/blob/56f8ead505c0c80bf92b0041e9e51f7f95964513/pygenometracks/tests/test_hiCMatrixTracks.py#L113-L213
In fact most of the ini files are written by the tests themselves. I kept only the ones I needed to run generateAllOutput.sh.

LeilyR:
Thanks!

Issue #47, opened 2018-10-26
Track for loops


Phlya:
Hi, would be great to have a new track type for showing loops on Hi-C data. For example, a bedpe file could be used to plot rectangles (and this could include plotting TADs, if the coordinate pairs are the same). Something like this: 
![image](https://user-images.githubusercontent.com/2895034/47573971-7d648080-d936-11e8-807f-a5168bcafcdb.png)

http://higlass.io/app/?config=AJQIPRFlREeqK2a1pciz_g

joachimwolff:
This feature is already in development and part of the branch 'longrangecontacts' of HiCExplorer. It is implemented as a `--loop` option for hicPlotMatrix. However, it is just a red circle and not a nice square.
This branch is changing a lot, therefore I don't recommend it to use it. Please be a bit more patient, I hope to finish it this year and than I will bring it to pyGenomeTracks too.

Phlya:
Of course, no worries, thank you! I was just thinking of adding a new option for the links track myself, actually, but I suppose I'll just wait for when you add it then.

Phlya:
BTW a little rough around the edges, but I actually made a custom track for now, and it seems to work fine.

```
class LoopTrack(tracksClass.GenomeTrack):
    SUPPORTED_ENDINGS = ['.bedpe']  # this is used by make_tracks_file to guess the type of track based on file name
    TRACK_TYPE = 'loops'
    OPTIONS_TXT = """
height = 3
title = Loops
color = blue
line width = 1
"""
    def __init__(self, *args, **kwarg):
        super(LoopTrack, self).__init__(*args, **kwarg)
        if 'line width' not in self.properties:
            self.properties['line width'] = 1
        if 'line style' not in self.properties:
            self.properties['line style'] = 'solid'
        self.max_height = None
        valid_intervals = 0
        intervals = []
        line_number = 0
        with open(self.properties['file'], 'r') as file_h:
            for line in file_h.readlines():
                line_number += 1
                if line.startswith(('browser', 'track', '#', 'chrom1')):
                    continue
                try:
                    chrom1, start1, end1, chrom2, start2, end2 = line.strip().split('\t')[:6]
                except Exception as detail:
                    self.log.error('File not valid. The format is chrom1 start1, end1, '
                                   'chrom2, start2, end2\nError: {}\n in line\n {}'.format(detail, line))
                    exit(1)

                try:
                    start1 = int(start1)
                    end1 = int(end1)
[...]

Issue #30, opened 2018-06-18
Arcs color scale based on score


gtrichard:
Hello @fidelram and @vivekbhr ,

I started using pyGenomeTracks and I really enjoy it, thanks for the hard work 👍 

Would it be possible to use the last column of test.arcs file (https://github.com/deeptools/pyGenomeTracks/blob/master/pygenometracks/tests/test_data/test.arcs) as a basis for a color scale of the arcs in order to display, for example, the interaction strength between 2 loci?

https://github.com/deeptools/pyGenomeTracks/blob/master/pygenometracks/tests/test_data/test.arcs

Best,
Gautier

DonStephano:
The newest version of pyGenomeTracks is very very useful and saves a lot of time!

@gtrichard s idea of the color-scale would is great and would be a cool feature. For us it would be also helpful to add a rgb-based-column so that each arc or a previously specified group of arcs (i.e. all interactions from a promotor) is plotted in a separate color (see example).

![test](https://user-images.githubusercontent.com/32867294/43714968-325c1d7e-997f-11e8-8bc8-bd30ece16763.png)

YichaoCaiDBS:
I am diving into the pyGenomeTracks and I found it very useful! Thanks a lot for the tool!

After digging into the enhancements that you are currently working on (#102), I found that the color-scale based on the score column has been added into the pull request.
Is it possible to add in the feature of using additional columns in .arc file for customization of the plotting style? That would be super useful.

For example,
```
# chrom1    start1    end1    chrom2    start2    end2    score    [additional column]
chr1    1    100    chr1    501    600    4    red
chr1    201    300    chr1    501    600    4    blue
```

Use the addition column as the attribute of the color when doing the plotting. Something like assigning colors to different levels of factors in R/ggplot2.

For example, if the 'color' field in .ini file specified as a color map name, scores are used to scale color to each arc (which has been included in the pull request #102 ). If the 'color' field in .ini file specified as 'predefined', the additional column would be used as the predefined color when calling `plot_arcs()`.

For further versatile plotting, would you consider including user pre-defined attributes if it is possible? For example, include a field of 'additionalColumn' in .ini file to be used as the key of the additional column to be used in `self.properties` that overrides the default value.
For example, in .arc file:
```
chr1    1    100    chr1    501    600    4    red    dotted
chr1    201    300    chr1    501    600    4    blue    dashed
```
In .ini file:
```
[test arcs]
file = test.arcs
[...]

Issue #103, opened 2019-10-01
Links track discrepancy between the help and the code


lldelisle:
Dear all,
I am working on the LinksTrack to use the score as color but I noticed:
In the help:
https://github.com/deeptools/pyGenomeTracks/blob/1352c2a16ab4ada31bcf8fa06d34df3a83422dad/pygenometracks/tracks/LinksTrack.py#L13-L15
In the code:
https://github.com/deeptools/pyGenomeTracks/blob/1352c2a16ab4ada31bcf8fa06d34df3a83422dad/pygenometracks/tracks/LinksTrack.py#L91-L92
I would rather use the middle but could you give me your opinion especially @fidelram who initially wrote it.

Many thanks,

Lucille

lldelisle:
`use_middle` is now available in version 3.5

PR #114, opened 2019-10-24
Links

lldelisle:
Enhancements:
- allow to plot the arcs of the links with a color scale based on scores as proposed in #30
- allow to plot rectangle on a heatmap for loops as discussed in #47 but included in the Links tracks (using links type = loops), thanks @Phlya .

Minor:
- bedpe is now a possible ending for Links track
- it is possible to have no score for this track
- put back the possibility to use the line_width to show the score in Links track.


lldelisle:
On this topic there is still the issue #103 for which I did not have any answer.

pygenometracks/tracks/LinksTrack.py
```diff
@@ -98,12 +121,55 @@ def __init__(self, *args, **kwarg):
...
+            self.properties['line width'] = 0.5
 
         if 'alpha' not in self.properties:
             self.properties['alpha'] = 0.8
 
+        if 'color' not in self.properties:
+            self.properties['color'] = DEFAULT_LINKS_COLOR
+        self.colormap = None
+        # check if the color given is a color map
+        if not matplotlib.colors.is_color_like(self.properties['color']):
```
bgruening: That entire color handling looks like a nice utility function that could be reused and maybe moved out of this python script.
But this can happen later ...

lldelisle: Yes, you are right it is also used more or less in the bedTrack. I should make a utility but if you don't mind I would do it when both #114 and #116 are merged. It will reduce the conflicts.

pygenometracks/tracks/LinksTrack.py
```diff
@@ -178,11 +274,54 @@ def plot_triangles(self, ax, interval):
...
         if y2 > self.max_height:
             self.max_height = y2
+
+    def plot_loops(self, ax, loop):
+        """
+              " <- 2
+        3->  "  " <- 1
+               " <- 0
+            """
+        from matplotlib.patches import Polygon
```
bgruening: I would move this to the top of the script.

pygenometracks/tracks/LinksTrack.py
```diff
@@ -156,7 +224,29 @@ def plot(self, ax, chrom_region, region_start, region_end):
         self.log.debug('title is {}'.format(self.properties['title']))
 
     def plot_y_axis(self, ax, plot_ax):
-        pass
+        if self.colormap is not None and self.properties['overlay previous'] == 'no':
+            import matplotlib.pyplot as plt
```
bgruening: I would move this to the top.

pygenometracks/tracks/LinksTrack.py
```diff
@@ -46,17 +53,25 @@ def __init__(self, *args, **kwarg):
         valid_intervals = 0
         interval_tree = {}
         line_number = 0
+        has_score = True
```
bgruening: general idea for later ... should we have functions that set the defaults? This way we can change default globally ...

pygenometracks/tracks/LinksTrack.py
bgruening: Agreed!

pygenometracks/tracks/LinksTrack.py
lldelisle: Yes I agree.

pygenometracks/tracks/LinksTrack.py
lldelisle: See #119

pygenometracks/tracks/LinksTrack.py
lldelisle: Done.

pygenometracks/tracks/LinksTrack.py
lldelisle: See #119

bgruening:
Merci!

Issue #55, opened 2018-12-19
`file_type = domains` has to be set when plotting triangles (version 2.0)


GinaRe:
`file_type = domains` has to be set when plotting triangles (version 2.0) showing TADs. This is missing in the documentation. Using the example from the documentation does not work.

fidelram:
Can you do a PR trying to fix this? That would be really helpful.

lldelisle:
Hi,
With #149, we can now use `display = triangles` or `file_type = domains`. They would both give the same result. Thank you for spotting it.

PR #116, opened 2019-10-24
Bedgtf

lldelisle:
Enhancements:
- added the possibility to merge transcripts into one single gene representation when using gtf.

Minors:
- added a warning when a boundaries_file is used.
- for the bed track, the value of display is now checked. It should be part of ['collapsed', 'triangles', 'interleaved', 'stacked', 'interlaced', 'domain'] with domain, interlaced, and interleaved all equivalent. This fix the issue #55.

We can discuss this last part, for me it was good to keep retro-compatibility but we can decide just one word and stick to it.

/!\ This might conflict with the PR #110
/!\ I used as new properties `merge transcripts` and `prefered name` but if you want I can change to `*_*`


pygenometracks/tracksClass.py
```diff
@@ -369,6 +369,10 @@ def check_file_exists(track_dict, tracks_path):
                 file_field_name = key
         # for file_field_name in ['boundaries_file', 'file', 'categories_file']:
         #     if file_field_name in track_dict:
+                if file_field_name == 'boundaries_file':
```
bgruening: Can you please add a comment for us, so we remove that line of code if we go towards 1.0.

pygenometracks/tracks/BedTrack.py
```diff
@@ -118,6 +126,19 @@ def __init__(self, *args, **kwarg):
             else:
                 self.colormap = self.properties['color']
 
+        # check the display is valid:
+        if self.properties['display'] in ['interlaced', 'domain']:
```
bgruening: can you define both valid/invalid ontop of the file after the imports with capital letters ... something like ... DISPLAY_BED_VALID = [...] and DISPLAY_BED_INVALID = [..] then we can reuse them and have a central place to change it....

pygenometracks/tracksClass.py
lldelisle: Done

pygenometracks/tracks/BedTrack.py
lldelisle: Done

bgruening:
@lldelisle a final rebase is needed here, sorry.

lldelisle:
No problem.

LeilyR:
could you please merge the current deeptools:develop to your branch

lldelisle:
This is done...

lldelisle:
But it was done before @bgruening  commented that's why you do not see it after the comments...

LeilyR:
sure

PR #118, opened 2019-10-25
Bedgraphbw

lldelisle:
Enhancements:
- added the possibility to rasterize bedgraph plots
- added the possibility to use summary functions on bedgraphs (creating a small temporary bigwig file) see #105 .
Minors:
- fixed warning from matplotlib when using fill for bw and bdg with different color for pos and neg (with the help of @Phlya ).
- try to avoid duplicated code
- fixed the error when using a bgz file which does not have index close to it.

A test with rasterize is missing. I need to write one.


lldelisle:
Test added.

lldelisle:
The test with the rasterize is skipped by travis...

lldelisle:
Now it is actually tested and worked.

PR #121, opened 2019-10-29
Boolean

lldelisle:
everywhere no=off=0=false and yes=on=1=true (case insensitive)
For me this is more light for the developer point of view.
I would like to write somewhere what are the one we should keep on the examples, for example:
```
GOOD_PRACTICES={'labels': ['on', 'off'],
                'show data range': ['yes', 'no'],
                'plot horizontal lines': ['yes', 'no'],
                'use middle': ['yes', 'no'],
                'rasterize': ['yes', 'no'],
                'global max row': ['yes', 'no'],
                'show_masked_bins': ['yes', 'no'],
                'show labels': ['yes', 'no'],
                'use summit': ['yes', 'no'],
                'skip': ['yes', 'no'],
                'merge transcripts': ['on', 'off']}
```
but I do not know where we should put it...


lldelisle:
Useless as #124 is more complete.

LeilyR:
what do you think if we always set them to them same thing for consistency? Like always yes or no

lldelisle:
What I like with the boolean is that when you code it is more light to write:
```
if self.properties['show_labels']:
```
Than
```
if self.properties['show_labels'] == 'yes':
```
But it is up to you. If you prefer to have yes/no everywhere, I can do it.

LeilyR:
> What I like with the boolean is that when you code it is more light to write:
> 
> ```
> if self.properties['show_labels']:
> ```
> 
> Than
> 
> ```
> if self.properties['show_labels'] == 'yes':
> ```
> 
> But it is up to you. If you prefer to have yes/no everywhere, I can do it.

No no, I definitely agree with you on that. It is of course a better way of coding. I was talking about the dictionary you have had in your first comment. I meant instead of having sometimes yes/no , some other times on/off in it, it would be better if we stick one form. Also I think there are more boolean objects to be added to that dictionary.

lldelisle:
Sorry you are talking about the best practices?
This is not used in this PR but will be used in the next one.

lldelisle:
The `GOOD_PRACTICES` dictionary is used in the next PR: #136 
I did not see any other boolean but if you find another one please tell me so I can update https://github.com/lldelisle/pyGenomeTracks/blob/boolean/pygenometracks/tracksClass.py#L347-L354

lldelisle:
@LeilyR , you were right, I was missing `nan_to_zeros`. It is now corrected.

lldelisle:
@bgruening, do you mean that 
1. we should warn people when they are not using true/false in this version but keep accepting no/yes/on/off and in the next 1.0 version only accept true/false?
or
2. Just use only true/false in our examples but keep accepting anything as this is a method from parser: `parser.getboolean(section_name, name)`?

bgruening:
My proposal would be:

* merge this one
* adopt examples as soon as we have time to just use true/false
* warn people when they are not using true/false in one of our next versions but keep accepting no/yes/on/off and in the next 1.0 version only accept true/false?

lldelisle:
OK for me this one can be merged in develop and I am correcting #136 to fit with the idea of using only 'true/false'.

PR #124, opened 2019-10-31
Added for each class a dictionnary with defaults and use them to initialize and add a config checker

lldelisle:
Summary of the PR:
- everywhere no=off=0=false and yes=on=1=true (case insensitive)
For me this is more light for the developer point of view.
The good practices keywords are present in the README.
- there are defaults defined on top of each track class that are used to initialize the self.properties.
- Before reading any file there is a checker of the config file.


lldelisle:
I added a python script to generate a markdown table for the possible options:
see https://github.com/lldelisle/pyGenomeTracks/blob/default/README.md#possible-parameters
I think this would be very useful.

bgruening:
Nice! Long term I think we should use https://pypi.org/project/configchecker or a similar library to validate configs ... or, and I guess this is very questionable, we could migrate to yaml - which has nice schemas and validation options.

lldelisle:
@bgruening I tried configchecker and I think it is not appropriated but you are right, we should implement a config checker.

lldelisle:
Do you prefer:
- I merge the ini_format branch in my default branch or 
- you first merge your ini_format into the develop and then I will solve the conflits?

LeilyR:
I need to fix some matplotlib issues in ini_format , that is why i havent merged it to dev yet. I am actually thinking what could be better. what if you just change the base of your pr to ini_format, then we can see the conflicts, if there is any? Or we can wait and merge this PR after I am done with ini_format (you second option). My comment was mainly a heads up to avoid any redundant work. Cheers!

bgruening:
@lldelisle no pressure to get the ini stuff in, right? Why not merging this one and you rebase the ini branch when you are ready?

lldelisle:
I would like to include a real config checker before merging to develop.

lldelisle:
I put a WIP on the title and I write when I am ready, it is ok for you?

lldelisle:
@LeilyR ,
maybe to avoid redundant work, you could open a PR with the {WIP] label. This way I am aware of what you are doing because I have no way to check if you create a branch and you are working on it.
For the moment, I had issues to check the differences between our branches because github see too many conflits. What we can do is:
1. I finish this PR.
2. You (@bgruening  and @LeilyR ) review it.
3. Merge it to the develop branch
4. I propose a PR that will solve the conflicts with your branch.

LeilyR:
It throws many conflicts because, unfortunately, there are. I have changed all double names to name1_name2 and you still have them all as name1 name2. It is even going to be too much work for you if want to correct them again. If you are not in rush for these changes please wait a little while, I wouldn't merge it now if it is not urgent.  Let's see what @bgruening  thinks.

lldelisle:
I am ok with waiting. There is no rush it is more a long term view.

lldelisle:
I am waiting for travis but it should be ready for review.

LeilyR:
So, shall we close this one?

lldelisle:
Yes

Issue #120, opened 2019-10-28
Behaviour of missing chromosome name inside a narrowPeak/begraph file


mevers:
## Versions 

pyGenomeTracks
```
pgt --version
pgt 3.1.2
```

Python
```
python --version
Python 3.7.4
```

## Issue

When I include a `narrowPeak` file that does not contain any entries for a specific chromosome (say `1`) and then plot a region involving that chromosome, the following error is produced

```
*Error*
Neither 1 or chr1 exits as a chromosome name inside the bedgraph file.
```

I'm not sure if this is intended behaviour but it would be great to have the option to have pyGenomeTracks produce an empty track when plotting a region for which there are no entries in a `narrowPeak` file, rather than throwing an error.

This would be good for two reasons:
1. We often look at specific genes/regions on chromosomes/scaffolds/contigs and want to show coverage and MACS2 `narrowPeak` tracks. However, the MACS2 `narrowPeak` file might not contain *any* peaks for that particular chromosome/scaffold/contig. We'd still like to show the coverage track and an empty `narrowPeak` track.
2. The error breaks our automated `snakemake` analysis pipeline, in which we generate pyGenomeTracks  plots for a set of different genes/regions. Having pyGenomeTracks produce an empty track for those genes/regions where there are no events in the `narrowPeak` file would allow the `snakemake` workflow to successfully finish without an error.

LeilyR:
Hi Maurits,
Thanks for bringing this issue up. It shouldn't be like that, I will have a look and will get back to you. May I ask if you could produce the same error using bed file tracks instead of bedgraphs?

mevers:
Hi @LeilyR.

The same error occurs if we include a simple BED file.

Here is a minimal reprex:

### `tracks.ini`
```
[x-axis]
where = top
title = Position
fontsize = 8

[bed test]
file = test.bed
```

### `test.bed`
```
1	150000000	180000000	sample_feature	0	.
```

Running
```
pgt --tracks tracks.ini --region 2:100000000-200000000 -out test.pdf
```

produces the following error
```

title not set for 'section 2. [bed test]'
INFO:pygenometracks.tracksClass:time initializing track(s):
INFO:pygenometracks.tracksClass:0.0010640621185302734
DEBUG:pygenometracks.tracksClass:Figure size in cm is 40 x 1.5. Dpi is set to 72

INFO:pygenometracks.tracksClass:plotting 1. [x-axis]
[...]

PR #125, opened 2019-10-31
solved issue with log

lldelisle:
Solved `AttributeError: 'NarrowPeakTrack' object has no attribute 'log'` of #120 .
In fact as bedgraphtrack did not have a self.log, all daughers tracks (bedgraphmatrixtrack, epilogostrack and narrowpeaktrack) were also impacted.


lldelisle:
Travis passed:
https://travis-ci.org/deeptools/pyGenomeTracks/builds/605461258?utm_medium=notification&utm_source=github_status
But it is still yellow...

PR #128, opened 2019-11-01
Narrow peak debug ld

lldelisle:
I changed the output of `get_scores` when there is data on the chromosome but not in the region to plot. The default is now `[], []`,
I added the check case in the vlines.
I added 2 tests.


bgruening:
That looks nice! Thanks.

PR #136, opened 2019-11-26
Initialize self.properties with default values

lldelisle:
Now, each track of GenomeTrack has 3 dictionnaries:
- `DEFAULTS_PROPERTIES` is a dictionary where each key is a parameter and each value is the default value when it is not set or when something goes wrong.
- `SYNONYMOUS_PROPERTIES` is a dictionary where each key is a parameter, each value is a dictionnary where each key is a string that should be replaced by the value (for example, `SYNONYMOUS_PROPERTIES = {'max_value': {'auto': None}}`)
- `POSSIBLE_PROPERTIES` is a dictionary where each key is a parameter, each value is an array with the only possible values for this parameter, if the value specified by the user is not part of them, it will be substituted by the default value.

The `__init__` of GenomeTracks calls `set_properties_defaults`

The `set_properties_defaults` of GenomeTracks do:
1. For each parameter which was not in self.properties, it puts the default value
1. For each parameter in `SYNONYMOUS_PROPERTIES` if the value is part of the dictionary it will substitute it by the value.
1. For each parameter in  `POSSIBLE_PROPERTIES`, it will check that the value in self.properties is part of the possible ones. If not, it will put the default value.

Now all tests such as: `if 'orientation' in self.properties` are useless as `self.properties` has been initialized.

A python script getAllDefaultsAndPossible.py uses the default properties and the possible properties to create a markdown table that can be integrated in the README as well as some text describing the possible options for some parameters.

The templates used by make_tracks_file were updated.

All the imports were put at the beginning of the file.

Everywhere, yes/no, on/off, True/False for boolean parameter was set to true/false and the error/warning were added in consequence.

I put the part of the README regarding the new tracks in a different README.

I put the treatment of type for coverage tracks (fill, line:0.5...) in a function in GenomeTracks (`process_type_for_coverage_track`).

I put the reading of a link file in a new method ('process_link_file`) of the LinksTrack class.

I corrected some space to _ that were missed.


pygenometracks/tracks/BedGraphTrack.py
```diff
@@ -216,6 +221,7 @@ def plot(self, ax, chrom_region, start_region, end_region):
             score_list, x_values = self.get_values_as_bdg(score_list,
                                                           pos_list)
 
+        # I do not see how this part of the code can be executed:
```
lldelisle: I think we should remove this part as I do not see how is it possible that self.properties['extra'] is an array...

lldelisle:
@LeilyR ,
The second bullet is ready for review. I put a long description of exactly what contains this PR at the beginning.

LeilyR:
What do you think if we make `true/false` case insensitive ?

bgruening:
```
def asbool(obj):
    if isinstance(obj, string_types):
        obj = obj.strip().lower()
        if obj in truthy:
            return True
        elif obj in falsy:
            return False
        else:
            raise ValueError("String is not true/false: %r" % obj)
    return bool(obj)
```
from https://github.com/galaxyproject/galaxy/blob/ed6f5d45a9ecbf0cf8a85341e0213b82a9132a46/lib/galaxy/util/__init__.py#L844

pygenometracks/tracksClass.py
```diff
@@ -356,10 +356,16 @@ def parse_tracks(self, tracks_file):
...
                         log.error("In section {}, {} was set to {}"
                                   " whereas we should have a boolean value."
-                                  "Please, use "
-                                  "{}".format(section_name, name, value,
-                                              str([k for k in parser.BOOLEAN_STATES])))
-                        exit()
+                                  "Please, use true or false."
+                                  "".format(section_name, name, value))
+                        exit(1)
+                    if value not in ['true', 'false']:
```
lldelisle: @LeilyR , you would change it to: `if value.lower() not in ['true', 'false']:`?

lldelisle:
To deal with boolean, in the previous PR, I used:
[getboolean from configparser](https://docs.python.org/3.6/library/configparser.html#configparser.ConfigParser.getboolean).
In this method, the true/false is case insensitive.
What I added in this PR is to add a warning if it was not true/false to warn the users that their config file may not work in the future.
But if you prefer, I can change the cases where I put the warning with:
> @LeilyR , you would change it to: `if value.lower() not in ['true', 'false']:`?

pygenometracks/tracksClass.py
LeilyR: yes, sounds like a good idea to me.

lldelisle: Done

pygenometracks/tracksClass.py
```diff
@@ -356,10 +356,16 @@ def parse_tracks(self, tracks_file):
...
+                        log.warning("Deprecation Warning: "
```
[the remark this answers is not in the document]
lldelisle: Indeed, you are right. It is changed.

lldelisle:
I think this PR is useless now. It was to split the big PR in pieces to help to review.

PR #141, opened 2019-12-04
changed the way introns arrow are plotted in UCSC style

lldelisle:
Hi,
People in my lab were not using the UCSC style because the arrows were too close one to the other.
I discovered that the arrows were plotted thanks to characters. For visual uniformity, I changed it to 2 lines with the color and the width of the backbone.
I also added the parameter arrow_interval to be able to make the arrows on the introns closer or more spaced.
It is true that the visual of tracks plotted in UCSC style will change but I think it is for better.


bgruening:
@lldelisle cool looks good for me. Nice improvement!

Issue #140, opened 2019-12-02
track label size


rikrdo89:
I noticed that my track labels (on the side) , specially the long ones get truncated and split in two or more lines. Is there a way to control the size of the labels and/or increase the space for track labels? I tried using the font_size argument but I think it only affects the label within the track itself. Thanks for your help!

lldelisle:
Hi,
You can control the fontsize of the titles of the tracks using the argument `--fontSize`:
```
pyGenomeTracks --tracks tracks.ini --region X:2000000-30000 -out tracks.png --fontSize 8
```
But this will not change where the titles are cut. You can also change the fraction of the width dedicated to titles: `--trackLabelFraction`.
See `pgt --help` to have more details.
Then depending if `--trackLabelFraction` is below 0.1 or above it, the title will be cut in lines of maximum 12 characters or maximum 30 characters.
This is written here:
https://github.com/deeptools/pyGenomeTracks/blob/1352c2a16ab4ada31bcf8fa06d34df3a83422dad/pygenometracks/tracksClass.py#L113-L116

So, if you were using the default parameters, you can use a wider  `--trackLabelFraction`, for example 0.1 and if needed, adjust the `--fontSize`. However, if you need lines longer than 30 characters, this is not possible for the moment.

It could be a good idea to adjust this number (12 or 30) dynamically using the fontsize and the  `--trackLabelFraction`. I will add it to my todo list. Thanks for reporting.

lldelisle:
Hi,
This has been solved in the new version #149 , the text is now adapted to the width available. So playing with `--trackLabelFraction` and `--fontSize` should allow you to display the titles the way you want.
If you still have issue, feel free to reopen it.

PR #145, opened 2019-12-06
Title adjust

lldelisle:
Adjust the split of the title of the tracks with the trackLabelFraction (see #140 ).
I just used the parameter `wrap=True` in matplotlib text...

After merging this PR, we need to regenerate all plots...


bgruening:
Ok, this now looks also good. Puh ... thanks a lot!

PR #156, opened 2019-12-12
Solve bug with boolean

lldelisle:
Hi,
I just noticed that if you put spaces in the name of a boolean it was failing giving a key error.
This PR solve the issue.
For example:
global max row = true
Will give a KeyError.

### DIFF
--- a/pygenometracks/tracksClass.py
+++ b/pygenometracks/tracksClass.py
@@ -419,6 +419,8 @@ class PlotTracks(object):
                              " parameter {} but there is no more parameter"
                              " with space in name. Will be substituted by {}."
                              "".format(section_name, old_name, name))
+                else:
+                    old_name = name
                 # end
                 SYNONYMOUS_PROPERTIES = track_class.SYNONYMOUS_PROPERTIES
                 # If the name is part of the synonymous we substitute by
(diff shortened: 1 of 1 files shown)


bgruening:
Do we need a point release?

lldelisle:
I don't know. I think it would be better because the error is really ununderstandable...

bgruening:
Do you want to do this? You just need to go to release and create a new one. The upload should happen automatically.

lldelisle:
Let's try...

lldelisle:
How do you put it into master? Do you make a PR from develop to master or there is a smarter way?

bgruening:
It does not need to be on master. You simply add a tag (on the most recent commit on master). You can do this via the GitHub interface.

lldelisle:
You mean the most recent commit on develop?

bgruening:
Oh I see, sorry, this PR went into develop. Ok, then develop needs to be merged into master before.
Bugs should be imho fixed on master. develop is for new stuff, maybe?

LeilyR:
this seems so strange to me that it failed for a space, where did we have `global max row` anyway? all phrases like that should have been changed to a connected phrase with `_`

LeilyR:
I first misunderstood you. I got how you mean now. I think this is how it should be right? It should have thrown that error. The spaced keys shouldn't get accepted. that behavior is  what it should have been

lldelisle:
Hi,
My idea was that as we were are doing a 0.1 release, for me this was tough for users to make all previous config file not working. So I put in version 3.2 already published:
https://github.com/deeptools/pyGenomeTracks/blob/422824d4199e22dea4b34b6862fa33dd3aba7cb7/pygenometracks/tracksClass.py#L414-L422
Which means that in the 3.2 version the space keywords are still working but there is a deprecation warning and the space keyword are not present anymore in none of our examples/tests/README.
When we will do a version 4. It will raise an error.
Then I realized that I was using:
https://github.com/deeptools/pyGenomeTracks/blob/422824d4199e22dea4b34b6862fa33dd3aba7cb7/pygenometracks/tracksClass.py#L433-L434
with the corrected name which may not be part of the parser.
So I just changed:
https://github.com/deeptools/pyGenomeTracks/blob/d25d321e0e9b519224205ac16c0e410806afe4f1/pygenometracks/tracksClass.py#L414-L424
and
https://github.com/deeptools/pyGenomeTracks/blob/d25d321e0e9b519224205ac16c0e410806afe4f1/pygenometracks/tracksClass.py#L436-L437
to make it work.
I am waiting for your green light before doing a 3.2.1.

bgruening:
Green from Grüning ;)

Issue #27, opened 2018-05-29
feature request: log transform the bigwig file


silvewheat:
It should be a common requirement to show expression level in a log-scaled y-axis.
My be adding a option to transform the y-axis can be considered.

lldelisle:
After nearly 2 years... This is in the version 3.4...

PR #157, opened 2019-12-12
Log scale

lldelisle:
To answer #27 
As it was opened 1 year and a half ago, no emergency...


pygenometracks/utilities.py
```diff
@@ -194,35 +194,36 @@ def plot_coverage(ax, x_values, score_list, plot_type, size, color,
...
+            raise Exception(msg)
         else:
-            warnings.warn('The transform: {} for file {} is not valid.'
-                          'will not use any transformation'.format(transform,
-                                                                   file))
-            return(score_list)
+            return(np.log(score_list + 1))
+    elif transform == '-log':
+        if np.nanmax(score_list.max) >= 0:
+            msg = ("\n*ERROR*\ncoverage contains positive or zero values.\n"
```
LeilyR: dont you think that coverage may have 0?

pygenometracks/tracks/BedGraphTrack.py
```diff
@@ -222,7 +224,10 @@ def plot(self, ax, chrom_region, start_region, end_region):
             score_list, x_values = self.get_values_as_bdg(score_list,
                                                           pos_list)
 
-        plot_coverage(ax, x_values, score_list, self.plot_type, self.size,
+        new_scores = transform(score_list, self.properties['transform'],
```
LeilyR: what about transformed_scores instead of new_scores? just to be more informative with the naming.

pygenometracks/utilities.py
lldelisle: Hi,
I don't know what should be done:
- option 1: We do not apply transformation when data are not in the correct range
- option 2: We put to nan the values that would give infinite values.
I used option 1 because this was what was done for HiC:
https://github.com/deeptools/pyGenomeTracks/blob/d25d321e0e9b519224205ac16c0e410806afe4f1/pygenometracks/tracks/HiCMatrixTrack.py#L117-L132
But we can decide to change both to option 2. No problem for me.

pygenometracks/tracks/BedGraphTrack.py
lldelisle: You are right, I will change it.

pygenometracks/utilities.py
LeilyR: you can allow for log2 or/and log10 with a user-defined pseudo count.

lldelisle: Yes, that's true. I was just thinking about doing the same as in HiC that's why I just put log/-log/log1p...

lldelisle: I think there is no point to use log2 or log10 as they are all proportional and the scale will still give the original value but I could add a log_pseudocount which is used with log and -log.
I think we still need to choose an option (raise an error or put nan).

LeilyR: we definitely need a pseudo count. 
How do you mean by there is `no use` and `still give the original value`? log2 means the values are transformed by numpy.log2 and log10 means they get transformed by numpy.log10, this can be decided by user.  How can these values be the same as the original ones?
why would it be any nan? If there is any, it can throw an error for now, but i cannot see that in the context of gene expression if you don't have 0.

lldelisle: > we definitely need a pseudo count.

There was log1p with a pseudo count of 1 but I will add log_pseudocount, no issue with that.

- Why did I say there was no difference between log2 and log10?
Here is an example of output with no transformation or with log1p:
master_log1p
The values on the y axis are not the log1p values but the original values. So if I replot the same example with log1p or log2(1+values) or log10(1+values) I will always have the same result because log2(1+values) = log(1+values)/log(2) and the same way log10(1+values) = log(1 + values)/log(10). I think the idea of the original issue was to have a a y-axis with log scale but maybe I misunderstood.

- It is possible to have tracks with only NA and no 0 values. For example, I have a 4Cseq track which is like this:
master_log
But I agree this is quite rare but you could have a track with ratio for which you would like to do a log without pseudocount.

LeilyR: >     * I think the idea of the original issue was to have a a y-axis with log scale but maybe I misunderstood.

yes that is what they have asked for.

LeilyR: >  a pseudo count of 1 but I will add log_pseudocount

+1 was an example, that is why i have suggested to make it a user define parameter with some default value.

LeilyR: I think it is a good idea to have log10 and log2 on top the natural log that you have already implemented or even instead of it.

lldelisle: If I understand well, you would prefer to put in y-axis the transformed value, not the original value, right?
I just noticed that it was depending of browser.
For example in igv keeps original values:
image
But UCSC put tranformed:
image
Let's put another parameter: y_axis_values = 'original' or 'transformed'.
And I add 'log2' 'log10' as possible values for transform
Would it be ok for you?

LeilyR: sure

lldelisle: Here is the new output:
master_log

joachimwolff:
Is this ready to merge?

lldelisle:
No, please do not merge this one. I just noticed something recently.

lldelisle:
Sorry, I had misunderstood the -log option in HiCMatrix. I thought it was a short cut for log(-values) but it is -log(values). I put it in coverage as in the HiCMatrix (-log(values)).

lldelisle:
@ LeilyR, I am wondering, should we raise error when the log cannot be apply or should we just put a message and replace by NA?
For info, in HiCMatrix, there is a section where it is supposed to check that but if there is a NA in the matrix the test is always false then the log is applied even with negative values...

lldelisle:
@LeilyR , I am wondering, should we raise error when the log cannot be apply or should we just put a message and replace by NA?
For info, in HiCMatrix, there is a section where it is supposed to check that but if there is a NA in the matrix the test is always false then the log is applied even with negative values...

LeilyR:
maybe it is better if it returns an error message and fail.

lldelisle:
OK I just added a test where applying log fails with a bigwig with 0 values.

lldelisle:
This is moved to #223

PR #158, opened 2019-12-13
Bug with boolean fix

lldelisle:


LeilyR:
@bgruening  there should be no old_name new_name in code. I think this needs to be reverted, what do you think? We had a long conversation about getting rid of all spaced name, i don't see any point to accept them again. Am I missing a point here?

PR #160, opened 2019-12-17
added a progress bar with tqdm

lldelisle:
Hi,
Some users complained that it was quite long. I thought that a progress bar for critical steps could be useful.
This involve a new dependency (tqdm) which is in conda/pip etc.
This might be slightly slower because you read twice bed files (once to count the number of line, the other one to guess the bed type, store intervals...).
But still I think it is useful.


bgruening:
This now only works for GTF and BED isn't it?

lldelisle:
It works for everything which uses file_to_intervaltree so bed, bedgraph, epilogos, narrowpeak and the gtf.

PR #162, opened 2019-12-17
Big files

lldelisle:
Hi,
Some people use pgt with big bedgraph or big gtf. This leads to long execution time. If a bedgraph or a gtf is larger than 10MB a warning is given suggesting to convert the bedgraph to bigwig or the gtf to bed. For this late suggestion, I implemented a new standalone binary which convert a gtf to bed using the same code as in readGtf.


pygenometracks/tracksClass.py
```diff
@@ -494,6 +495,30 @@ def parse_tracks(self, tracks_file):
             # The track_options will be checked for the file paths:
             track_options = self.check_file_exists(track_options,
                                                    tracks_file_path)
+
+            # If a big bedgraph is used a warning is printed:
+            if track_class == BedGraphTrack:
+                file_size = os.path.getsize(track_options['file'])
```
bgruening: This could be a function with max_file_size and maybe a more general error message?

lldelisle: I think it should not be an error but a warning. What I like with the 2 different messages here it that they give a solution (bedgraph -> bigwig, gtf -> bed). A more general message would loose these advise. But I agree, this could be in the README and we could use a general message to refer to it. I will do that.

bgruening: Na then keep it. I just thought that a function could reuse some code. maybe that is overdoing it ... leave it like it is if you prefer :)

lldelisle:
If it is merged and #172 also, the message should be adapted and display only `if pRegion is None`.

lldelisle:
Not sure it is really relevent now. Let close it and see how it goes.

Issue #51, opened 2018-10-31
pyGenomeTracks & Epilogo's for a >8kb region


favilaco:
Hi,

I'm having issues when trying to plot more than 8kb regions.

This is my input.qcat

```
chr10	12829801	12830001	id:1174730,qcat:[ [-0.2403,25], [-0.004582,17], [0,13], [0,24], [0,23], [0,22], [0,21], [0,20], [0,19], [0,16], [0,14], [0,1], [0,12], [0,11], [0,10], [0,9], [0,6], [0,5], [0,4], [0,3], [0,2], [0.004148,15], [0.009821,18], [0.2731,8], [0.5698,7] ]
chr10	12830001	12830201	id:1174731,qcat:[ [-0.2858,25], [0,13], [0,24], [0,23], [0,22], [0,21], [0,20], [0,18], [0,16], [0,15], [0,1], [0,12], [0,10], [0,9], [0,6], [0,5], [0,4], [0,3], [0,2], [0.006585,17], [0.00834,14], [0.04288,11], [0.1326,19], [0.2277,8], [0.5698,7] ]
chr10	12830201	12830401	id:1174732,qcat:[ [-0.2638,25], [0,13], [0,24], [0,23], [0,22], [0,21], [0,20], [0,16], [0,1], [0,12], [0,10], [0,9], [0,6], [0,5], [0,4], [0,3], [0,2], [0.004148,15], [0.006585,17], [0.00834,14], [0.009821,18], [0.01357,11], [0.0322,7], [0.04267,19], [0.7562,8] ]
chr10	12830401	12830601	id:1174733,qcat:[ [-0.198,25], [0,13], [0,24], [0,23], [0,22], [0,21], [0,20], [0,16], [0,1], [0,12], [0,10], [0,9], [0,6], [0,5], [0,4], [0,3], [0,2], [0.001742,19], [0.00834,14], [0.009821,18], [0.01357,11], [0.0237,17], [0.0322,7], [0.1122,15], [0.4468,8] ]
chr10	12830601	12830801	id:1174734,qcat:[ [-0.1028,25], [0,13], [0,24], [0,23], [0,22], [0,21], [0,20], [0,18], [0,16], [0,1], [0,12], [0,10], [0,9], [0,6], [0,5], [0,4], [0,3], [0,2], [0.001742,19], [0.00834,14], [0.01357,11], [0.0171,7], [0.04988,15], [0.06851,17], [0.2502,8] ]
[...]

PR #163, opened 2019-12-17
fixed cases with no data in qcat file

lldelisle:
Relative to #51 
Maybe I need to change the formulation. A hole means a region with no corresponding line in the qcat file.


pygenometracks/tracks/EpilogosTrack.py
```diff
@@ -90,6 +90,10 @@ def plot(self, ax, chrom_region, start_region, end_region):
             # the qcat_json is a pseudo json line, that misses
             # the { } and the quotes. The following lines fix that.
             qcat_json = qcat_json[0]
+            if not isinstance(qcat_json, str):
+                # This would happen if the qcat file has a hole.
```
bgruening: What about:

```suggestion
                # This would happen if the qcat file has a missing value.
                #The missing value is filled with np.repeat(np.nan, ..) and should be skipped here.
```

PR #165, opened 2019-12-18
Dif / ratio / any operation on the fly for coverage tracks

lldelisle:
Following #32 
I implemented operations between coverage tracks on the fly.
There are now 2 new parameters for bigwig and bedgraph: `second_file` and `operation`.
~~When `second_file` is set,~~ the `operation` is computed.
```
# to compute operations on the fly between 2 bigwig files
#second_file = path for the second file
# operation will be evaluated, it should contains file and second_file,
# we advice to use nans_to_zeros = true to avoid unexpected nan values
#operation = file - second_file
#operation = log2((1 + file) / (1 + second_file))
#operation = max(file, second_file)
```
- Possible improvements:
  - ~~Maybe the `operation` could be computed even if there is no `second_file`, I do not know if this could be useful to normalize a dataset (for example, currently, `operation = file * 0.89` will not work but `operation = second_file * 0.89` would work).~~
  - ~~Currently, if one of the two files has no value in the region, even if `nans_to_zeros = true` there will be no plot. This should probably be changed.~~


lldelisle:
Also I noticed that it is more painful than useful to have Epilogs, NarrowPeak, and BedGraphMatrix as children of BedGraph. What do you think of a BedPlusTrack, daughter of GenomeTrack and BedGraph, Epilogs, NarrowPeak, and BedGraphMatrix would be daughters of BedPlusTrack?
With a scheme:
```
Curently:                                    Could be:
           Genome                                      Genome
             |                                           |
          BedGraph                            ______  BedPlus ____
         /   |     \                         /       |      |     \
bdgMatrix Epilogos NarrowPeaks           bdg   bdgMatrix Epilogos NarrowPeaks
```

lldelisle:
To be able to scale a file or do an operation between 2 files:
```
# to compute operations on the fly on the file
# or between 2 bigwig files
# operation will be evaluated, it should contains file or
# file and second_file,
# we advice to use nans_to_zeros = true to avoid unexpected nan values
#operation = 0.89 * file
#operation = - file
#operation = file - second_file
#operation = log2((1 + file) / (1 + second_file))
#operation = max(file, second_file)
#second_file = path for the second file
```

bgruening:
Can we also point users to `bigwigCompare` from deeptools. On the fly is nice, but I assume there is still a value in doing the per-processing before if you doing a lot of plots?

pygenometracks/tracks/BedGraphMatrixTrack.py
```diff
@@ -68,6 +70,23 @@ class BedGraphMatrixTrack(BedGraphTrack):
     INTEGER_PROPERTIES = {}
     # The color cannot be set for the moment
 
+    def __init__(self, properties_dict):
+        GenomeTrack.__init__(self, properties_dict)
+
+        self.tbx = None
```
bgruening: can this move up one level in the class hierarchy as a optional function?

lldelisle: Thanks good idea

lldelisle:
> Can we also point users to `bigwigCompare` from deeptools. On the fly is nice, but I assume there is still a value in doing the per-processing before if you doing a lot of plots?

Where would you do that? In the docs? In the `make_tracks_file` output?

lldelisle:
> > Can we also point users to `bigwigCompare` from deeptools. On the fly is nice, but I assume there is still a value in doing the per-processing before if you doing a lot of plots?
> 
> Where would you do that? In the docs? In the `make_tracks_file` output?

@bgruening

bgruening:
mh, normally we put this into the release notes.

`pyGenomeTracks/docs/content/releases/3.4_announce.rst` could be one spot? Together with all new features? I guess this would also be useful later?

bgruening:
What about:

making a `*` behind the `second_file` and `operation` and adding a sentence under the table. Something like: `While pyGenomeTracks can convert coverage tracks on the fly, this might be a time-consuming step, especially on large files and if you want to replot many times. In this situation, we recommend using the deepTools suite to convert your files in advance. For example [bamCoverage](https://deeptools.readthedocs.io/en/develop/content/tools/bamCoverage.html)`

lldelisle:
This is moved to #223

PR #168, opened 2020-01-08
added alpha to the FLOAT_PROPERTIES in Links

lldelisle:
Solve a small bug: alpha cannot be used in links tracks because I forgot to add it to FLOAT_PROPERTIES.


bgruening:
Do we want to make a point release here?

lldelisle:
No, I do not think it is needed.

PR #174, opened 2020-01-10
useBetterChromMapping

lldelisle:
To be able to deal with changes in chromosome name.
I propose here an option.
Using https://github.com/dpryan79/ChromosomeMappings.git
I generate a big dictionary with chromosome names as key and a list with all possible synonymous chromosome names which are different from the key.
I store it in _chromMapping.py (I think this is not optimal, the file is 12MB, probably it slows down).
When we want to plot on a chromosome which is not on the list of available chromosome we check if there is any of the synonymous which is available and we use it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ jobs:
     install:
     - conda install --yes python=$TRAVIS_PYTHON_VERSION flake8
     script:
-    - if [[ "$LINT" == "1" ]]; then flake8 . --exclude=.venv,.build,build --ignore=E501,F403,E402,F999,F405,E712,W503
+    - if [[ "$LINT" == "1" ]]; then flake8 . --exclude=.venv,.build,build,_chromMapping.py --ignore=E501,F403,E402,F999,F405,E712,W503
       ; fi
 before_install:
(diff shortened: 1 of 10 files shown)


LeilyR:
I think a better solution would be leave be as it used to be, I don't think one program is supposed to handle everything. I don't see anything wrong with failing with a meaningful error message. Users can then fix their format and get back to it. If you want we can update the documentation about that. The only reason you go through it is that if they have a matrix with ucsc format genome and they ask for a region in ensemble format right? If so, why not they just rerun it with the right region format? Is there any other reason for this change that I am missing?

lldelisle:
If you have files in different format, for example, you have a bed file in ensembl format and a bedgraph in UCSC.
If you want to plot on regular chromosome using +/- chr is working.
If you want to plot on mitochondria or contigs... Then you will have to modify your input files so they fit into the same nomenclature.
This is not only for HiC, this is for all tracks. For the moment, nobody report such a case so this is maybe too much to integrate it.

LeilyR:
I really appreciate your effort but i do think it may not worth adding 12mb of data to the repo. regardless how even the mapping procedure would perform later.

lldelisle:
OK

PR #183, opened 2020-01-29
fix cases when whole coding region is in a single intron

lldelisle:
This is a bug fix (that's why I PR to master).
When an exon had both UTR start and UTR end within it, only the UTR start was considered:
master_bed_arrow_zoom_before
I fixed it in both UCSC and flybase style (have a look to ng1 to ng4):
master_bed_arrow_zoom


LeilyR:
what was the base of your branch? master or dev?

lldelisle:
master

LeilyR:
but you also changed the test files, right? this is not going to be so good since develop would still have the old test files. Do you mind if we still merge it into dev? Is it really necessary to have it out now? I suppose we are gonna tag a release in a month or 2. Is it possible for your users to use develop for the time being?

lldelisle:
No problem to merge it into develop.

PR #189, opened 2020-02-12
Fix empty bed

### LINKED ISSUE #181: Crashing if empty input for TAD boundaries file
Hi,

using the current version and plotting a Hi-C matrix and Tad boundaries. If now the TAD boundaries are empty, pyGenomeTracks is crashing without any useful error message.

I see two options: a) do not plot any Tad boundaries because file is empty but not let the whole thing crash or b) check all inputs if they contain data and if not return error message which input is actually empty.

Best,

Joachim

lldelisle:
Fixes #181


lldelisle:
For the moment if the matrix is empty it stills stop. 
If you ask for log/log1p and you have negative values with no NA it also stops.
Should we change it?

PR #203, opened 2020-04-01
Type = gtf

lldelisle:
This was in the help of the track when using make_tracks_file but not in the code. This is now in the code.


pygenometracks/plotTracks.py
```diff
@@ -150,15 +150,7 @@
 from pygenometracks._version import __version__
 from .utilities import InputError
 
-DEFAULT_BED_COLOR = '#1f78b4'
```
LeilyR, line 153: why did you remove these defaults?

lldelisle: Because they are not used here so I thought this would be confusing.

lldelisle:
Just to make it clear, this option is only useful when the gtf file does not end with gtf or gtf.gz...

LeilyR:
you mean otherwise it is more efficient to treat them as a bed file? If so, I wonder if it is a good idea to have this or we can easily guide users to rename their files.

pygenometracks/plotTracks.py
LeilyR: Ok I see, it is because you have them set for each of the arguments that you have modified, right?

lldelisle: All the default colors/colormap are defined on top of their class, the DEFAULT_TRACK_HEIGHT and DEFAULT_MARGINS are defined in tracksClass.py. This is like this since I started to work on PGT... I just decided to remove the defaults from places where they are not used... (This was done in #198 but has been merged to master not dev...).

lldelisle:
No, it is impossible to read it as bed:
In gtf, first column is chr, start is 4th, end is 5th, strand is 7th... This is very different.
I just mean that as soon as your file ends with `.gtf` or `.gtf.gz` you don't need to set type = gtf but you need it if you have a file which ends with `.dat` like in galaxy...
In the test file `bed_and_gtf_tracks.ini`, you have one example where the file extension is good:
```
[test gtf]
file = dm3_subset_BDGP5.78.gtf.gz
height = 10
title = gtf from ensembl
fontsize = 12
file_type = bed
```
and one where it is not:
```
[test gtf collapsed]
file = dm3_subset_BDGP5.78_gtf.dat
height = 10
title = gtf from ensembl; extension is dat; type=gtf one entry per gene
merge_transcripts = true
prefered_name = gene_name
fontsize = 12
type = gtf
file_type = bed
```
The second one requires type=gtf.

lldelisle:
I finally decided to remove it from the help and create a new file_type following @LeilyR advice.

lldelisle:
See #215

PR #204, opened 2020-04-02
Decreasing xaxis

lldelisle:
To be able to compare data on different species it can be useful to use a decreasing Xaxis.
I added it as an argument of the main command: `--decreasingXAxis`, it will put the xaxis from region end to region start instead of region start to region end. The most difficult part was with the labels of the bed, to keep them on the right...
I introduced some tests to confirm I got the expected output and @aurHin will test it.


pygenometracks/plotTracks.py
```diff
@@ -224,6 +224,12 @@ def parse_arguments(args=None):
                         default='left',
                         choices=['left', 'right', 'center'])
 
+    parser.add_argument('--decreasingXAxis',
+                        help='By default, the x-axis is increasing but if you '
+                             'want to see all the tracks with a decreasing X '
+                             'axis. Put this option.',
```
bgruening, line 230:
```suggestion
                         help='By default, the x-axis is increasing.'
                            'Use this option if you want to see all tracks with a decreasing x-axis.',
```

pygenometracks/plotTracks.py
```diff
@@ -296,8 +304,10 @@ def main(args=None):
             sys.stderr.write("saving {}\n".format(file_name))
             # print("{} {} {}".format(chrom, start, end))
             trp.plot(file_name, chrom, start, end, title=args.title,
-                     h_align_titles=args.trackLabelHAlign)
+                     h_align_titles=args.trackLabelHAlign,
+                     decreasingXAxis=args.decreasingXAxis)
```
bgruening, line 308: the variable should be `decreasing_x_axis`, maybe?

pygenometracks/tracks/BedTrack.py
```diff
@@ -419,40 +421,73 @@ def plot(self, ax, chrom_region, start_region, end_region):
                 self.counter += 1
                 bed = region.data
 
+                if ax.get_xlim()[0] < ax.get_xlim()[1]:
+                    bed_left = bed.start
+                    bed_right = bed.end
+
+                    def add_to_right(a, b):
+                        return(a + b)
```
bgruening, line 429: the brackets are not needed

pygenometracks/tracks/BedTrack.py
```diff
@@ -419,40 +421,73 @@ def plot(self, ax, chrom_region, start_region, end_region):
...
-                    row_last_position.append(bed_extended_end)
+                    row_last_position.append(bed_extended_right)
                 else:
                     # If all_labels_inside = True
                     # genes which goes over will have their labels inside
                     if self.properties['all_labels_inside'] and self.properties['labels'] \
-                       and bed_extended_end > end_region:
-                        start_pos = int(bed.start - (num_name_characters * self.len_w))
+                       and is_right_to(bed_extended_right, ax.get_xlim()[1]):
+                        bed_extended_left = int(add_to_left(bed_left, (num_name_characters * self.len_w)))
```
bgruening, line 473: add_to_left is now always an int? So the cast can be removed?

pygenometracks/tracksClass.py
```diff
@@ -202,7 +202,7 @@ def get_tracks_height(self, start_region=None, end_region=None):
         return track_height
 
     def plot(self, file_name, chrom, start, end, title=None,
-             h_align_titles='left'):
+             h_align_titles='left', decreasingXAxis=False):
```
bgruening, line 205:
```suggestion
             h_align_titles='left', decreasing_x_axis=False):
```

pygenometracks/tracksClass.py
```diff
@@ -267,7 +267,10 @@ def plot(self, file_name, chrom, start, end, title=None,
                     width_inch = label_axis.get_window_extent().width
                     width_dpi = width_inch * self.dpi / fig.dpi
 
-            plot_axis.set_xlim(start, end)
+            if decreasingXAxis:
```
bgruening, line 270:
```suggestion
            if decreasing_x_axis:
```

pygenometracks/plotTracks.py
lldelisle: Thanks it is better.

pygenometracks/plotTracks.py
lldelisle: You are right.

pygenometracks/tracks/BedTrack.py
lldelisle: done

pygenometracks/tracks/BedTrack.py
lldelisle: No add_to_left is just doing addition or substraction but I need the cast.

PR #215, opened 2020-05-04
[WIP] file_type = gtf

lldelisle:
To make it less confusing now gtf has its own file type.
gtf in file_type = bed are only supported until the next 1.0.


pygenometracks/tracks/GtfTrack.py
```diff
@@ -0,0 +1,155 @@
...
+import numpy as np
+
+DEFAULT_BED_COLOR = '#1f78b4'
+DISPLAY_BED_VALID = ['collapsed', 'triangles', 'interleaved', 'stacked']
+DISPLAY_BED_SYNONYMOUS = {'interlaced': 'interleaved', 'domain': 'interleaved'}
+DEFAULT_DISPLAY_BED = 'stacked'
+
+
+class GtfTrack(BedTrack):
+    SUPPORTED_ENDINGS = ['gtf', 'gtf.gz']
```
LeilyR, line 14: does it matter? I thought you wanted to check for the format of the file based on the given data type and check if the format is right regardless of the ending. Did I get you wrong?

lldelisle: The SUPPORTED_ENDINGS is used by make_tracks_file

lldelisle:
I still need to include the tests when a bed is put with file_type = gtf. I think it should raise an error. For the contrary, I am working on the bed format on PR #210 but I will test how it is currently working.

PR #216, opened 2020-05-05
fix url for test_data

### LINKED ISSUE #211: tests/test_data not available - trying to determine data format for chromatin track color designation
Hi, I noticed that the test data is no longer available.  I was hoping to do something similar to the "chromatinStates_kc.bed.gz" and be able to designate specific colors to the specific tracks.  Is this possible?

lldelisle:
Fix #211

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ jobs:
     - LINT=1
     - TRAVIS_OS_NAME=linux
     before_install:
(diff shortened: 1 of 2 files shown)


LeilyR:
thanks for the fix! do you mind if it goes through the dev branch? Our dev branch is falling way behind the mater. I will merge it immediately to master before merging any other PR and tag a bug fix release.

PR #221, opened 2020-05-05
New version of file_type = gtf

lldelisle:


lldelisle:
To make it less confusing now gtf has its own file type.
gtf in file_type = bed are only supported until the next 1.0.
In the file_type = bed, the first line is read and from it you can guess the type of bed it is (BED6, BED8, BED9 or BED12). If you have less than 6 valid fields (you need at least 3 but the rest can be filled with default values).
I added a test:
If you put a bed file with file_type = gtf you get an input error: "This is not a gtf file" and if you put a gtf file (without gtf or gtf.gz extension) with file_type = bed you get an input error: "This is probably not a bed file".

pygenometracks/tracks/BedTrack.py
```diff
@@ -987,7 +998,9 @@ def _plot_small_arrow(self, ax, xpos, ypos, strand):
         :
         :return: None
         """
-        if strand == '+':
+        if strand == '.':
+            return()
```
bgruening, line 1002: just "return"

pygenometracks/readBed.py
```diff
@@ -176,27 +139,17 @@ def get_bed_interval(self, bed_line):
...
-            # To be able to process it as bed6, the extra-fields are removed.
-            line_data = line_data[:6]
-
-        # If the file_type is below, values will be added.
+        if not is_first_line:
+            if self.file_type != 'bed6':
+                # When bed6 you can have less fields in one row
+                # because there are default values
+                assert len(line_data) >= self.fields_to_read, \
+                    "File type detected is {} but line {}: {} does " \
```
bgruening, line 147: Maybe we should require python >=3.6 and use f-strings :)

pygenometracks/readBed.py
```diff
@@ -176,27 +139,17 @@ def get_bed_interval(self, bed_line):
...
+                    "not have {} fields.".format(self.file_type,
+                                                 self.line_number,
+                                                 bed_line,
+                                                 self.fields_to_read)
```
bgruening, line 151:
```suggestion
                assert len(line_data) >= self.fields_to_read, \
                    f"File type detected is {self.file_typ} but line {self.line_number}: {bed_line} does " \
                    f"not have {self.fields_to_read} fields.")
```

bgruening: This should now be possible from 3.6 onwards.

pygenometracks/readBed.py
```diff
@@ -221,31 +174,77 @@ def get_bed_interval(self, bed_line):
...
                 try:
                     line_values.append(int(r))
                 except ValueError:
-                    sys.stderr.write("Value: {} in field {} at line {}"
-                                     " is not an integer"
+                    raise InputError("Value: {} in field {} at line {}"
+                                     " is not an integer. This is "
+                                     "probably not a bed file."
                                      "\n".format(r, idx + 1,
                                                  self.line_number))
```
bgruening, line 186:
```
                    raise InputError(f"Value: {r} in field {idx + 1} at line {self.line_number}"
                                     " is not an integer. This is "
                                     "probably not a bed file.\n")
```

pygenometracks/tracks/BedTrack.py
lldelisle: done

lldelisle:
I think we should do a global reformatting of all code if we want to use f-Strings so I would put it into a different PR.

lldelisle:
> @lldelisle can we add the deprecation warning also to the docs?

Where could we put it? In a new section?

bgruening:
mh, normally we put this into the release notes.

`pyGenomeTracks/docs/content/releases/3.4_announce.rst` could be one spot? Together with all new features? I guess this would also be useful later?

lldelisle:
OK

PR #222, opened 2020-05-05
remove hack for older python versions

bgruening:
This should fix the conda install during testing. Seems that the URL to minconda is now a redirect.
It also removes an old hack from our setup.py file.


lldelisle:
I think this should go to develop (We already fixed the conda curl -L in develop but this should not make a conflict).

bgruening:
So we should leave master broken? I can point this to develop if you like.

lldelisle:
We are going to make a 3.4 today that's why it is better to put it into develop so the merge in master will be easier if we fix all the conflicts now...

bgruening:
Have targetted the develop branch now.

PR #223, opened 2020-05-05
Log scale + Operation

lldelisle:
Combination of #157 and #165 :
Allow to make operations in the fly as proposed in #32 using 2 parameters: `operation` and `second_file`. Here is the corresponding documentation:
```
# to compute operations on the fly on the file
# or between 2 bigwig files
# operation will be evaluated, it should contains file or
# file and second_file,
# we advice to use nans_to_zeros = true to avoid unexpected nan values
#operation = 0.89 * file
#operation = - file
#operation = file - second_file
#operation = log2((1 + file) / (1 + second_file))
#operation = max(file, second_file)
#second_file = path for the second file
```
Allow to put a log scale as proposed in #27 using 2 parameters: `transform` and `pseudocount`.
Here is the corresponding documentation:
```
# To log transform your data you can also use transform and log_pseudocount:
# For the transform values:
# 'log1p': transformed_values = log(1 + initial_values)
# 'log': transformed_values = log(log_pseudocount + initial_values)
# 'log2': transformed_values = log2(log_pseudocount + initial_values)
# 'log10': transformed_values = log10(log_pseudocount + initial_values)
# '-log': transformed_values = - log(log_pseudocount + initial_values)
# For example:
#tranform = log
#log_pseudocount = 2
```
When you use this transformation, the y axis will write the transformation applied, alternatively, you can decide to keep on the y axis the original values with:
```
# When a transformation is applied, by default the y axis
# gives the transformed values, if you prefer to see
# the original values:
#y_axis_values = original
```
Here is an example of the similarities/differences you would get between operation and transform:
master_log


pygenometracks/tracks/BigWigTrack.py
```diff
@@ -73,6 +108,14 @@ class BigWigTrack(GenomeTrack):
...
         super(self.__class__, self).__init__(*args, **kwargs)
         self.bw = pyBigWig.open(self.properties['file'])
+        self.bw2 = None
+        if 'second_file' in self.properties['operation']:
+            if self.properties['second_file'] is None:
+                raise InputError("operation: {} requires to set the parameter"
+                                 " second_file."
+                                 "".format(self.properties['operation']))
+            else:
+                self.bw2 = pyBigWig.open(self.properties['second_file'])
```
bgruening, line 118: We probably need to add a destructor like `def __del__(self):` that closes the file handle properly.

lldelisle: Solved.

PR #229, opened 2020-05-06
[WIP] Release 3.4

lldelisle:
Here is the text I propose for the release annoucement (following merge of new documentation for each track):

Improved documentation:

- fix url of test_data thanks to @sonnynguyen
- Better documentation for pyGenomeTracks / makeTracksFile
- Documentation of each track

Enhancements:

- Different bed files format (BED3, BED4, BED5, BED6, BED8, BED9, BED12) is guessed based on the first line and rely not only on number of columns but on the type of each column (string/float etc.). This way, broadPeaks/gappedPeaks and other bed-like formats can be accepted as `file_type = bed` without preprocessing.
- gtf has now its own `file_type`. **Warning**:`file_type = bed` for gtf will be no longer supported in a future version.
- You can use a decreasing x-axis (from larger coordinates to smaller coordinates) with the option `--decreasingXAxis` in `pyGenomeTracks`.
- For bedgraph and bigwig:
  - logScale: To apply a log transformation to your data you can set `transform` to `log1p`, `log`, `log2`, `log10`, `-log`. For others than `log1p`, you can set a `log_pseudocount`. By default the y axis values will represent the transformed values and the transformation will be written on the left of the y axis but you can decide to keep the original values by setting `y_axis_values = original`.
  - operation on files: You can plot the result of an operation on one or two files without preprocessing. To use it, put the operation in `operation` with the variable `file` or both variables `file` and `second_file`, for example `operation = 0.89 * file` or `operation = max(file, second_file)` or `operation = file - second_file`. In the two later cases, you need to put the path of the second file in `second_file`. However, this operation can be time consuming if you do it multiple times and you may prefer to convert your files in advance.
For more examples on these two new features, visit [the examples](https://pygenometracks.readthedocs.io/en/latest/content/examples.html#log-transform-and-operation-examples)
- Both `type = line` and `type = lines` are working in bedgraph_matrix

Bugfix:

- In version 3.3, if you were using a colormap in bed the `border_color` and `color_utr` were ignored. Now each one can be set to different color/colormap/bed_rgb.


lldelisle:
Please do not merge before I merge a PR with the release text for readsthedoc.

lldelisle:
I think we need to decide what to do with pip:
- We do not support pip anymore -> We need to write it in the release note + in the doc.
- We put HiCMatrix 12 on pip
- We relax the dependency on HiCMatrix (to 11) but we write in the documentation that pip will be no longer supported in future versions...

joachimwolff:
We literally just submitted a paper with the statement of pip support, so I guess I need to make HiCMatrix available for pip.

joachimwolff:
hicmatrix in version 12 is added to pypi.

lldelisle:
If #230 is accepted:


Improved documentation:

- fix url of test_data thanks to @sonnynguyen
- Better documentation for pyGenomeTracks / makeTracksFile
- Documentation of each track

Enhancements:

- Different bed files format (BED3, BED4, BED5, BED6, BED8, BED9, BED12) is guessed based on the first line and rely not only on number of columns but on the type of each column (string/float etc.). This way, broadPeaks/gappedPeaks and other bed-like formats can be accepted as `file_type = bed` without preprocessing.
- gtf has now its own `file_type`. **Warning**:`file_type = bed` for gtf will be no longer supported in a future version.
- You can use a decreasing x-axis (from larger coordinates to smaller coordinates) with the option `--decreasingXAxis` in `pyGenomeTracks`.
- For bedgraph and bigwig:
  - logScale: To apply a log transformation to your data you can set `transform` to `log1p`, `log`, `log2`, `log10`, `-log`. For others than `log1p`, you can set a `log_pseudocount`. By default the y axis values will represent the transformed values and the transformation will be written on the left of the y axis but you can decide to keep the original values by setting `y_axis_values = original`.
  - operation on files: You can plot the result of an operation on one or two files without preprocessing. To use it, put the operation in `operation` with the variable `file` or both variables `file` and `second_file`, for example `operation = 0.89 * file` or `operation = max(file, second_file)` or `operation = file - second_file`. In the two later cases, you need to put the path of the second file in `second_file`. However, this operation can be time consuming if you do it multiple times and you may prefer to convert your files in advance.
For more examples on these two new features, visit [the examples](https://pygenometracks.readthedocs.io/en/latest/content/examples.html#log-transform-and-operation-examples)
- Both `type = line` and `type = lines` are working in bedgraph_matrix

Bugfix:

- In version 3.3, if you were using a colormap in bed the `border_color` and `color_utr` were ignored. Now each one can be set to different color/colormap/bed_rgb.

lldelisle:
Maybe, the `file_type` of `XAxisTrack` should be `x_axis` instead of `x-axis`.

pygenometracks/getAllDefaultsAndPossible.py
```diff
@@ -66,16 +83,20 @@ def main():
...
     mat[0, 0] = 'parameter'
     mat[1, 0] = '--'
     for j, track_type in enumerate(all_tracks_with_default, start=1):
-        mat[0, j] = track_type
+        mat[0, j] = f":doc:`tracks/{track_type}`"
         mat[1, j] = '-'
         for i, p in enumerate(all_default_parameters):
             if j == 1:
-                mat[i + 2, 0] = p
+                if p in putStarsAfter:
```
LeilyR, line 95: why?

pygenometracks/getAllDefaultsAndPossible.py
```diff
@@ -66,16 +83,20 @@ def main():
...
+                    mat[i + 2, 0] = p + "*"
```
LeilyR, line 96: and the same question here

pygenometracks/readBed.py
```diff
@@ -221,31 +174,77 @@ def get_bed_interval(self, bed_line):
                         r = '.'
                 line_values.append(r)
 
-            elif idx in [1, 2, 6, 7, 9]:
-                # start and end fields must be integers, same for thichStart(6),
-                # and thickEnd(7) and blockCount(9) fields
+            elif idx in [1, 2]:
```
LeilyR, line 177: why don't you need to check the others columns which were previously there?

pygenometracks/getAllDefaultsAndPossible.py
lldelisle: I put a star after second_file and operation to encourage people to preprocess their data...

lldelisle: This was proposed by Bjorn at the end of https://github.com/deeptools/pyGenomeTracks/pull/165

pygenometracks/readBed.py
lldelisle: I check them later: 1 and 2 are mandatory to be integer, if they are not I raise an error, whereas 6, 7, 9 are optional.

pygenometracks/getAllDefaultsAndPossible.py
LeilyR: I see

PR #236, opened 2020-05-18
added faq with HiCMatrix 12 to 13

lldelisle:
Add a notification in the documentation for HiCMatrix.

### DIFF
--- a/docs/content/faq.rst
+++ b/docs/content/faq.rst
@@ -1,0 +1,15 @@
+FAQ
+===
+
+.. contents:: 
+    :local:
+
+Why the scale of my Hi-C plot suddenly changed
+----------------------------------------------
+pyGenomeTracks is using `HiCMatrix <https://github.com/deeptools/HiCMatrix>`_ to read the matrix from ``h5`` and ``cool`` format.
+From version 12 to version 13, a normalization step when reading ``cool`` file was removed. This normalization was mostly used 
+when you were providing ``cool`` file from `cooler balance <https://cooler.readthedocs.io/en/latest/cli.html#cooler-balance>`_.
+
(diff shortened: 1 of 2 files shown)


docs/content/faq.rst
```diff
@@ -0,0 +1,15 @@
...
+    :local:
+
+Why the scale of my Hi-C plot suddenly changed
+----------------------------------------------
+pyGenomeTracks is using `HiCMatrix <https://github.com/deeptools/HiCMatrix>`_ to read the matrix from ``h5`` and ``cool`` format.
+From version 12 to version 13, a normalization step when reading ``cool`` file was removed. This normalization was mostly used 
+when you were providing ``cool`` file from `cooler balance <https://cooler.readthedocs.io/en/latest/cli.html#cooler-balance>`_.
+
+If you want to keep the old scale you need to downgrade to HiCMatrix version 12 but version 13 also correct some bugs so we advice
+to change your ``max_value`` in your parameter file to adjust to the new scale.
```
bgruening, line 14:
```suggestions
If you want to keep the old scale please adopt your ``max_value`` in your parameter file to the new scale.
```

bgruening: Do we need to mention the downgrade option? In the long run this is not an option either way.

lldelisle: Yes but if people have all figures of a paper done and just want to add one figure...

lldelisle: Maybe we can remove this proposition in few months...

lldelisle: If people have all figures of a paper ready and just want to update one plot I understand that they would like just to downgrade and we can remove this downgrade option in few months.

bgruening: I need to test this in Bioconda. I changed the suggestion ... is that any better?

Issue #214, opened 2020-05-03
Display y-axis ticks and grid lines in bigwig track graph


guerler:
It would be nice to enable y-axis ticks and automated horizontal grid lines in the bigwig track graph. Currently only the minimum and maximum values are shown by default which makes it difficult to read/guess the actual values.

![y-axis-not-informative](https://user-images.githubusercontent.com/2105447/80927461-1108df80-8d6c-11ea-9050-5ffa7a2ec28b.png)

lldelisle:
Hi,
Good improvement idea. We put it in the TODO list.

lldelisle:
Is it what you were looking for?
![master_grid](https://user-images.githubusercontent.com/30404086/82354029-8af1c780-9a00-11ea-8bba-827799e52476.png)

dannon:
@lldelisle That looks great, it would help a lot!  Is the line spacing configurable?

lldelisle:
No...
For the moment I only added major ticks

guerler:
Thanks @lldelisle, that is a huge improvement for the readability.

lldelisle:
For the moment, it is available here: https://github.com/lldelisle/pyGenomeTracks/tree/grid
But I am still working on it especially on the y axis when the log transformation is used.
It should be ready by tomorrow.

dannon:
@lldelisle Fantastic, thanks!

lldelisle:
Hi,
grid is now available in version 3.5.

PR #237, opened 2020-05-19
New parameter of bigwig and bedgraph: grid

lldelisle:
solve #214


bgruening:
@lldelisle magic ... thanks a lot.

PR #240, opened 2020-05-20
Links fix

lldelisle:
Solving #103 
I added use_middle parameter.


bgruening:
Is use_center maybe a better name?

lldelisle:
You are right. But I already used `use_middle` for bedgraph so I think it can be confusing if now I use `use_center` for links, no?

bgruening:
Ok, that's a good argument :)

PR #248, opened 2020-06-09
Solve bug when depth or plotting region is smaller than bin size

lldelisle:


LeilyR:
Is this really a bug you think? Do you think we should support that? I was wondering if this should throw an error message and fails. Don't see a point in plotting less than a bin, specially for the depth.

lldelisle:
When you want to display different resolution on different loci, you want to have a single ini file for all and for me it is good that pgt does not raise errors when one track is not valid...

LeilyR:
well if that is the case , maybe it can throw a warning message that the depth is smaller than a bin size and it needs to be increased to at least one bin and only plot the valid ones.

lldelisle:
OK I proposed it in #251

PR #252, opened 2020-07-07
Fix type error in python3.8

lldelisle:


lldelisle:
Either we make a minor release with just that (so I change the version and we merge to master) or we do a big release with all what is in the dev branch... What are your opinions (please do not merge before we agree)?

bgruening:
Ah big one :)

LeilyR:
I believe this PR cause complication in the installation process, I am afraid we should postpone it until bioconda support 3.8.
@bgruening Do you have any opinion on that?

lldelisle:
This PR is just bug fixes which were not a problem in python3.6 and python3.7 and become problems in python3.8 but the correction of the bugs should be merged I think.

lldelisle:
This PR does not allow to install pygenometracks with python3.8 on conda... For this we will wait for bioconda to support 3.8.

bgruening:
Bioconda will not support soon python3.8. But this also does not matter I think. The reviewer has not used Conda - I assume conda would have worked, as he would have get python3.7.

Should we not enable tests for python 3.8? https://github.com/deeptools/pyGenomeTracks/blob/master/.travis.yml#L20

lldelisle:
The dependencies (from bioconda) are not compatible with python3.8 on conda, so we would need to install them with pip... This is dirty, no?

joachimwolff:
I think the best approach is to have the changes for python 3.8 already in the src, we can test them local on our computers. However, because bioconda does not support 3.8 we don't test 3.8 on travis and we do not make the statement to the public that we support 3.8. 
As soon as bioconda supports 3.8 we can add the 3.8 support very fast because in the best case only the test cases on travis need to pass and we would be done.

lldelisle:
Or should we add tests for pip in all python version in travis?

lldelisle:
Can we have 2 travis files? One for conda and one for pip ?

joachimwolff:
I see bioconda as our main distribution and pip only as a side channel. I would not put all the potential work into a special pip support.

lldelisle:
Or use github ci for pip and keep travis for conda?

lldelisle:
> I see bioconda as our main distribution and pip only as a side channel. I would not put all the potential work into a special pip support.

OK

bgruening:
> The dependencies (from bioconda) are not compatible with python3.8 on conda, so we would need to install them with pip... This is dirty, no?

No, this is totally fine I think. We either way need to use `pip install` to install PGT.

> Can we have 2 travis files? One for conda and one for pip ?

No, but we can have many different tests in parallel in one travis file.

> I see bioconda as our main distribution and pip only as a side channel. I would not put all the potential work into a special pip support.

Maybe. But this does not mean people will not use pip. And pip or pypi is and will ever be the main python package installation procedure. If PGT is on pypi it should work. I think the problem currently is/was that https://pypi.org/project/pyGenomeTracks/ lists the requirement as `Requires: Python >=3.6.*, <4 ` which is due to https://github.com/deeptools/pyGenomeTracks/blob/master/setup.py#L127

I do think its worthwhile to test against all most recent python version. If pip is working conda will work too. We still can recommend clearly to use conda.

joachimwolff:
As I wrote: I think we should merge the 3.8 changes, test it locally on our computers. But as long as bioconda is not on 3.8 we don't make the statement that we support 3.8 at all. If someone wants to work with it and has no issues: Great! If they complain: We can point them to the statement we don't support it. 

I have especially our lovely matplotlib dependency in mind. In my experience always good for a surprise.

bgruening:
>  test it locally on our computers.

why not on CI?

> But as long as bioconda is not on 3.8 we don't make the statement that we support 3.8 at all.

This will not prevent users from doing it and complain. Its fine to not officially support it, but why not testing it against 3.8 in the first place and make sure it works?

PR #258, opened 2020-07-09
Solve bed bugs

lldelisle:
While increasing the coverage I found some bugs in readBed. I think it never happened to anyone but just in case...


lldelisle:
Ready for review

PR #259, opened 2020-07-10
Not require sorted bed

lldelisle:
In fact intervalTree sort the interval we give to it so no need to check that the bed is sorted before using it...


lldelisle:
Ready for review

PR #260, opened 2020-07-10
only accept chr:start-end as region input

lldelisle:
I realized it was not useful to plot on a whole chromosome when we don't know the size of it so I decided to be more stringent on the region provided by the user.


lldelisle:
Ready for review

PR #261, opened 2020-07-10
Solvemlogbug

lldelisle:
When I implemented the -log in PR #157 I left a `.max` which should not be here. I guess not too many users are using `-log` but again a bug found when increasing the number of tests...


lldelisle:
Ready for review

pygenometracks/utilities.py
```diff
@@ -256,7 +256,7 @@ def transform(score_list, transform, log_pseudocount, file):
         else:
             return(np.log1p(score_list))
     elif transform == '-log':
-        if np.nanmax(score_list.max) <= - log_pseudocount:
+        if np.nanmax(score_list) <= - log_pseudocount:
```
bgruening, line 259: oh this makes somehow sense :) Great catch!

PR #262, opened 2020-07-10
correct ylim

lldelisle:
Better adjust ylim for bed when gene_rows or global_max_row is set


lldelisle:
ready for review (travis is green but connection with github is bad)...

PR #266, opened 2020-07-14
Solve some bugs

lldelisle:
Major:
- line_width for vlines was ignored
- in coverage when using -log was not tested correctly for potential log of negative values

Minor:
- better error handling with custom color
- more stringent on bed12 (like in UCSC)
- do not show a warning when using `file_type = x_axis`
- only plot scale bar when in the region

### DIFF
--- a/pygenometracks/utilities.py
+++ b/pygenometracks/utilities.py
@@ -261,7 +261,7 @@ def transform(score_list, transform, log_pseudocount, file):
         else:
             return(np.log1p(score_list))
     elif transform == '-log':
-        if np.nanmax(score_list) <= - log_pseudocount:
+        if np.nanmin(score_list) <= - log_pseudocount:
             msg = ("\n*ERROR*\ncoverage contains values smaller or equal to"
                    f" - {log_pseudocount}.\n"
                    f"- log( {log_pseudocount} + <values>) transformation can "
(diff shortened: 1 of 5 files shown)


lldelisle:
This one is green on travis on my repo: https://travis-ci.com/github/lldelisle/pyGenomeTracks/builds/175550530 but I cannot relaunch the linting step here...

pygenometracks/utilities.py
```diff
@@ -261,7 +261,7 @@ def transform(score_list, transform, log_pseudocount, file):
         else:
             return(np.log1p(score_list))
     elif transform == '-log':
-        if np.nanmax(score_list) <= - log_pseudocount:
```
LeilyR, line 264: why did you change the max to min?

LeilyR: this should stay max, i would say. it is enough only if one is below this threshold.

lldelisle: No if one is below but not others, the max will be above.

lldelisle: Now I am confused... I thought we don't want to apply log to negative values. So if one value is below we don't want to compute, so if the min is below, no?

LeilyR: I understood that at the moment we cannot accept any negative value so you are giht and we can change that to min.

Issue #273, opened 2020-08-12
Issue producing images with BED/GTF tracks


egmg726:
Using: Python 3.6.2 and pyGenomeTracks 3.5 

I'm having issues visualizing BED and GTF files, but I am not getting any error messages when I am trying to make an image. It will simply fail to produce a file. However, I have no issues producing files when using only BigWig files. I have tried this out with my own data and the test data on your GitHub page. (The examples below use files from the test_data folder in this repo)

For instance, if I use this command (with no BED or GTF files):

<pre><code>$ pyGenomeTracks --tracks alpha.ini --region X:2700000-3100000 --trackLabelFraction 0.2 --dpi 130 -o master_alpha.png</code></pre>
I will get this output (and an image):

<pre><code>INFO:pygenometracks.tracksClass:initialize 1. [test bigwig]
INFO:pygenometracks.tracksClass:initialize 2. [test bigwig max]
INFO:pygenometracks.tracksClass:initialize 3. [test bigwig mean]
INFO:pygenometracks.tracksClass:initialize 4. [spacer]
INFO:pygenometracks.tracksClass:initialize 5. [test bigwig]
INFO:pygenometracks.tracksClass:initialize 6. [test bigwig max]
INFO:pygenometracks.tracksClass:initialize 7. [test bigwig mean]
INFO:pygenometracks.tracksClass:initialize 8. [spacer]
INFO:pygenometracks.tracksClass:initialize 9. [test bigwig]
INFO:pygenometracks.tracksClass:initialize 10. [test bigwig mean]
INFO:pygenometracks.tracksClass:initialize 11. [test bigwig min]
INFO:pygenometracks.tracksClass:initialize 12. [x-axis]
INFO:pygenometracks.tracksClass:time initializing track(s):
INFO:pygenometracks.tracksClass:0.007485389709472656
DEBUG:pygenometracks.tracksClass:Figure size in cm is 40 x 25.398936170212767. Dpi is set to 130

INFO:pygenometracks.tracksClass:plotting 1. [test bigwig]
INFO:pygenometracks.tracksClass:plotting 2. [test bigwig max]
INFO:pygenometracks.tracksClass:plotting 3. [test bigwig mean]
INFO:pygenometracks.tracksClass:plotting 4. [spacer]
INFO:pygenometracks.tracksClass:plotting 5. [test bigwig]
INFO:pygenometracks.tracksClass:plotting 6. [test bigwig max]
[...]

PR #274, opened 2020-08-24
add message if bedtools is not installed.

lldelisle:
Following #273 
I don't really like the print but I did not find another way.

### DIFF
--- a/pygenometracks/utilities.py
+++ b/pygenometracks/utilities.py
@@ -89,6 +89,9 @@ def temp_file_from_intersect(file_name, plot_regions=None, around_region=0):
         except NotImplementedError:
             print("It seems that BEDTools is not installed. Cannot subset the file.")
             file_to_open = file_name
+        except Exception as e:
+            print(e)
+            file_to_open = file_name
         sys.stderr.close()
         sys.stderr = sys.__stderr__
         with open(temporary_file.name, 'r') as f:
--- a/pygenometracks/utilities.py
+++ b/pygenometracks/utilities.py
@@ -86,6 +86,9 @@ def temp_file_from_intersect(file_name, plot_regions=None, around_region=0):
             file_to_open = original_file.intersect(regions, wa=True, u=True).fn
         except pybedtools.helpers.BEDToolsError:
             file_to_open = file_name
+        except NotImplementedError:
+            print("It seems that BEDTools is not installed. Cannot subset the file.")
+            file_to_open = file_name
         sys.stderr.close()
         sys.stderr = sys.__stderr__
         with open(temporary_file.name, 'r') as f:


lldelisle:
Should I change print to:
```
sys.__stderr__.write("It seems that BEDTools is not installed. Cannot subset the file.")
```
The other option is to put log as parameter of `file_to_intervaltree` and `temp_file_from_intersect` and then use `log.warn`.
What do you think?

joachimwolff:
How about this approach: 

- if all dependencies are present should be checked at installation time. I wonder a bit why it is not covered from our setup.py. There is already a part: https://github.com/deeptools/pyGenomeTracks/blob/master/setup.py#L93
- as additional checks: why don't we check it at every run of pgt at the start of the main function? And even more important, simply check if all dependencies are present, if not, throw an error and exit.

- I prefere to use `log` everywhere, similar to how it is used in HiCExplorer. I would do: 

```
log.error('dependency is missing')
exit(1)
```

lldelisle:
So, you would not allow it to work when BEDTools is not present?
My idea was just to warn the user that it will be longer but you are probably right.

lldelisle:
Hi,
I think the problem is that it is not a python dependency (pybedtools is the python dependency) but I don't know how to write an external dependency in the setup.py.
I don't like `exit(1)` because it is impossible to catch when you want to use it within python. I propose a more stringent version here: #284

joachimwolff:
Conda has bedtools as a dependency of pybedtools.

- How is it with pip? 

The setup is just a python script, we could simply call pybedtools like you do in the source code. If it is successful we can continue with the installation, if not, we interrupt it.

lldelisle:
pip also requires pybedtools, the problem is that pybedtools is working without bedtools, except if you call a bedtools command...
I know this is strange... You can create `pybedtools.BedTool` objects without bedtools but then when you do `object1.intersect(object2)` it fails... So to test it it is not just a import.

joachimwolff:
We can test if the binary exists, similar to this solution:
https://stackoverflow.com/questions/377017/test-if-executable-exists-in-python

lldelisle:
I prefer the #284 option just raise an error when you get a `notImplementedError` with a meaningful message.

lldelisle:
#284 would give as output (only when bedtools is required):
```
$ bin/pgt --tracks ./pygenometracks/tests/test_data/bedgraph_useMid.ini --BED ./pygenometracks/tests/test_data/regions_imbricated_chr2.bed --trackLabelFraction 0.2 --width 38 --dpi 130  -o ./pygenometracks/tests/test_data/master_bedgraph_useMid.png
INFO:pygenometracks.tracksClass:initialize 1. [x-axis]
INFO:pygenometracks.tracksClass:initialize 2. [spacer]
INFO:pygenometracks.tracksClass:initialize 3. [test bedgraph]
ERROR:pygenometracks.utilities:BEDTools is not installed but required.
$
```

joachimwolff:
We should give the users at least at the installation time a warning that not all dependencies are present. Maybe with the hint to use conda to avoid such trouble.

lldelisle:
Hi,
Back on this question. I did not managed to make a test to be sure bedtools is installed at the time of installation before pybedtools is installed...
If someone wants to do it, no problem but for me, raising an error when you need it is sufficient.
If you want to mention conda I can add a message in the error like:
```
ERROR:pygenometracks.utilities:BEDTools is not installed but required, installation with conda would avoid this type of error.
```

joachimwolff (approved):

Issue #278, opened 2020-08-31
No PDF document generated


SMorrison42:
**Welcome to the pyGenomeTracks GitHub repository! Before opening the issue please check
 that the following requirements are met :**

  - [ ] Search whether this issue (or a similar issue) has been solved before
  using the search tab above. Link the previous issue if appropriate below.

  - [ ] Paste your pyGenomeTracks version (`pyGenomeTracks --version`) and your python
  version (`python --version`) below.

pyGenomeTracks verison 3.5

python version 3.7

  - [ ] Paste the full pyGenomeTracks command that produces the issue below
  (ignore if you simply spotted the issue in the code/documentation).
pyGenomeTracks --tracks tracks.ini --BED test.bed -o test.pdf

  - [ ] Paste the output printed on screen from the command that produces the issue
  below (ignore if you simply spotted the issue in the code/documentation).
INFO:pygenometracks.tracksClass:initialize 1. [x-axis]
INFO:pygenometracks.tracksClass:initialize 2. [spacer]
INFO:pygenometracks.tracksClass:initialize 3. [test]

No PDF file is generated after the pyGenomeTracks command is generated

lldelisle:
Hi,
Thanks for the report.
It may be that bedtools is not accessible in your environment like in #273 
If bedtools was accessible, please could you give more details on the ini file and the file in the [test] section?

yanchunzhang:
> Hi,
> Thanks for the report.
> It may be that bedtools is not accessible in your environment like in #273
> If bedtools was accessible, please could you give more details on the ini file and the file in the [test] section?

This is very helpful! I got the same problem and was a little disappointed until I saw this answer. Could you please tell users to load bedtools in their environment before running pyGenomeTracks in your documentation or tutorial page? I think that'll help other users like me and the questioner.

lldelisle:
It will be solved in the next version:
You would have a message: 
"It seems that BEDTools is not installed. Cannot subset the file." I can also add it to the documentation.

lldelisle:
Hi,
The new version 3.5.1 take care of this.

Issue #277, opened 2020-08-31
summary_method: Invalid interval bounds


BenoitM-I2BC:
```
(python 3.8)
(pyGenomeTracks 3.5)
```

Hi, 
I report here what seems to be a bug in BedGraph Summary_Method. The error occurs when --region right border argument is **not** included in the bedgraph intervals. 
The left border of the --region argument does not have to be included in the bedgraph intervals.

Below is a minimal test file to (hopefully) reproduce to the bug.
All the instructions work without summary_method.

Best, 
Benoit

```
===== Bedgraph.bg ===== (the interval [400-404] is not included)
chr7	100	200	1
chr7	200	300	2
chr7	300	350	3
chr7	350	399	4
chr7	405	450	5
```

```
===== Track1.ini =====
[bedgraph]
file = Bedgraph.bg
height = 5
summary_method = max
file_type = bedgraph
```

```
### WORKS
pyGenomeTracks -o test1.png --tracks track1.ini --region chr7:100-399
pyGenomeTracks -o test1.png --tracks track1.ini --region chr7:100-406
pyGenomeTracks -o test1.png --tracks track1.ini --region chr7:50-399

```
```
### FAILS
pyGenomeTracks -o test1.png --tracks track1.ini --region chr7:100-400
pyGenomeTracks -o test1.png --tracks track1.ini --region chr7:100-401

### it returns:
Traceback (most recent call last):
  File "/usr/local/bin/pyGenomeTracks", line 11, in <module>
    main(args)
  File "/usr/local/lib/python3.8/site-packages/pygenometracks/plotTracks.py", line 331, in main
    trp.plot(args.outFileName, *regions[0], title=args.title,
  File "/usr/local/lib/python3.8/site-packages/pygenometracks/tracksClass.py", line 265, in plot
    track.plot(plot_axis, chrom, start, end)
  File "/usr/local/lib/python3.8/site-packages/pygenometracks/tracks/BedGraphTrack.py", line 320, in plot
    score_list, x_values = self.get_values_as_bigwig(score_list,
  File "/usr/local/lib/python3.8/site-packages/pygenometracks/tracks/BedGraphTrack.py", line 422, in get_values_as_bigwig
    scores_per_bin = np.array(bw.stats(chrom_region,
RuntimeError: Invalid interval bounds!

```

lldelisle:
Hi,
Good catch. I had not thought about this case. I made a pull request to solve this issue. It will be solved in the next release.

[...]

Issue #13, opened 2018-03-15
bedgraph loading is very slow


Phlya:
Hi, I understand that bigWigs should be preferred to bedgraphs for high resolution data, and they should be much faster than bedgraphs, but currently it seems that loading bedgraphs is unreasonably slow... A ~360Mb file takes a few seconds to load with pandas, but I have been waiting a few minutes for it to load as a `BedGraphTrack` and just had to interrupt it because I got bored. Is it something that might get improved? Thanks!

fidelram:
Honestly, takes less time to convert the bedgraph to bigwig than to wait each time. Pandas has super optimized parsers that can read tables quite fast that's the big difference. Moreover, pyGenomeTracks needs to create an interval tree of the bedgraph file to be able to index it which also adds some further time.

Other solution, besides the bigwig idea, is to split your bedgraph per chromosome. In the case of a mammalian size genome, that should reduce the loading time by about 20 times (e.g. 1 min vs. 20).

Phlya:
Awesome, thanks!

BenoitM-I2BC:
Hi, I recently discovered pyGenomeTracks and I’m very pleased to start using. 
This thread is rather old but my question is related.

The tabix support for bedgraph indeed allows to speed-up the loading of the data. 
Yet, the initialization of the track (before the loading bar appears) is rather slow. A 800Mb tabix-indexed bedgraph can take more than 1min to initialize (and 2-3 sec to retrieve the data).

Is this a normal behavior? Am I missing something obvious?
Thanks.

```
====== track.ini file =====
[test bedgraph tabix]
file = Data.bg.gz
title = tabix-indexed bedgraph
type = fill
summary_method = max
number_of_bins = 3000
file_type = bedgraph
===========================
```
```
(python 3.8)
(pyGenomeTracks 3.5)
```

lldelisle:
Hi,
First of all, we are happy to have new users...
For the moment, the code only consider a bedgraph as a potential tabix file if it ends with `.bgz`. So, If I am not wrong the fact that you tabix indexed your file is not used.
To test this, you can simply change your file name (don't forget to change the name of the index also) and see if it is quicker.
Since version 3.5, we do an intersection (with bedtools) between the bedgraph and the regions to plot, so it should be quicker than before. Can you give use the output you get during the initialization (especially the progress bar)?

Thanks

BenoitM-I2BC:
Hi,
Awesome ! that's what I was missing. 
[...]

PR #290, opened 2020-10-07
Update version 3.5.1

lldelisle:
Let say we wait for the end of the week for comments/identification of new bugs and then we release.


lldelisle:
The release text would be:

## Bugfixes:

- Get a message when bedtools is installed instead of crashing without any message (see #273, #278).

- Always test if a bedgraph is tabix indexed without checking the extension (see #13 )

- Fix a bug which was happening when ``operation`` or ``summary_method`` was used on bedgraph whereas the bedgraph had some missing intervals (see #277).

- Enforcing version 15 of HiCMatrix. Version 14 had a bug concerning the application of the correction factors of cool files.

LeilyR (approved):

PR #291, opened 2020-10-07
Abstract classes for bed and bedgraph

lldelisle:
Now the structure is:
```
- GenomeTracks
| - BigWigTrack
| - HiCMatrixTrack
| - HLinesTrack
| - LinksTrack
| - ScaleBarTrack
| - SpacerTrack
| - XaxisTrack
| - BedGraphLikeTrack
| | - BedGraphTrack
| | - BedGraphMatrixTrack
| | - EpilogosTrack
| | - NarrowPeakTrack
| - BedLikeTrack
| | - TADsTrack
| | - BedGeneLikeTrack
| | | - BedTrack
| | | - GtfTrack
```

This induced 1 change for the user which is that in the output of `make_tracks_file` it is not written:
```
# If the bed file contains the exon
# structure (bed 12) then this is plotted. Otherwise
# a region **with direction** is plotted.
```

Also I think we could decide that the TADsTrack has a 'normal' `plot_label` which means put the title vertically centered. This can be done easily by moving the `def plot_label` block from `BedLikeTrack` to `BedGeneLikeTrack`. Tell me if I do it or if I leave it like this.


joachimwolff:
Already mergable?

lldelisle:
Well, the only thing is if it is relevant to keep the `plot_label` different from all other tracks for the TAD track...
For the genes track I understand because we want to see where it starts but for the TADs...
We can also merge it and change it after...

PR #292, opened 2020-10-09
Remove dpi in tracks

lldelisle:
Hi,
(This PR has been built on top of abstract_classes this is why there are so many changes.)

The idea of this PR is to remove the need of DPI inside the `plot_label` method of each track class to be able to save an image in different resolution through API.
While I was doing this PR I noticed that the figures had pretty large margin on left and right and that both 'titles' and 'y-axis' were in fact printed in the margin:
master_hlines

When I removed DPI from the tracks, I forced the labels (titles) to fit in the axis:
master_hlines
 
Should we also make the 'y-axis' fit into its axis or do we leave it go to the margin.

Of course, as you can notice, this PR will change the proportion each column relative to the page so I need to update all outputs used in tests but before doing it I wanted to see with you the 'y-axis' and decide together the left and right margins.


joachimwolff:
How will the result look if you force the y-axis labels?

lldelisle:
We cannot really force the 'y-axis' because nothing is preventing matplotlib to write outside of the figure but with some changes in the code we could have:
master_arcs_no_score
master_hlines
But if you have very high values:
master_hlines2
The thing we could do is like we do for `--trackLabelFraction`, have a `--yAxisFraction` which would give:
master_hlines2
When set to 0.1...

PR #296, opened 2020-10-22
fix #295

### LINKED ISSUE #295: Bug: pink is not recognized as colormap
I need to check why and if other colormap are impacted. I guess it is because pink is also the name of a color...

lldelisle:
update process_color to deal with names which match both colormap and color


pygenometracks/tracks/GenomeTrack.py
```diff
@@ -326,37 +326,36 @@ def process_color(self, param, colormap_possible=False,
...
-                # For example:
-                # colormap = ['white', (1, 0.88, 2./3), (1, 0.74, 0.25), (1, 0.5, 0), (1, 0.19, 0), (0.74, 0, 0), (0.35, 0, 0)]
-                if self.properties[param][0] == '[':
+            # If someone what to use its own colormap,
+            # he can specify the rgb values or color values:
+            # For example:
+            # colormap = ['white', (1, 0.88, 2./3), (1, 0.74, 0.25), (1, 0.5, 0), (1, 0.19, 0), (0.74, 0, 0), (0.35, 0, 0)]
+            if self.properties[param][0] == '[':
+                try:
+                    custom_colors = eval(self.properties[param])
```
joachimwolff, line 335: For local usage I would accept it, however, eval allows that external code can be introduced and execute which is a no-go for servers like Galaxy. @bgruening what is your opinion on this function?

joachimwolff: @LeilyR any thoughts on the eval function?

lldelisle: As it is not directly linked to the PR itself, I moved it to #308

joachimwolff (approved):

PR #298, opened 2020-10-26
Fix labels in colorbar which sometimes are not corresponding to real values

### LINKED ISSUE #297: Bug: when fontsize is changed colorbar scale may be wrong
Here is an example of what can happen when we change the fontsize:
Left is without specifying the fontsize, Right is using `--fontSize 16`
image
You can see that the picture stay the same but the colorbar is wrong.
This is due to:
https://github.com/deeptools/pyGenomeTracks/blob/1c6b592d151abb217a2c4b0d6f5d828819dc9471/pygenometracks/tracks/HiCMatrixTrack.py#L378-L389
Because the label are fixed but not the ticks...

lldelisle:
Fix #297


pygenometracks/tests/test_data/browser_tracks_hic_force_scale.ini
```diff
@@ -0,0 +1,27 @@
...
+file_type = hic_matrix
+show_masked_bins = false
+
+[spacer]
+height = 0.5
+
+[hic matrix]
+file = Li_et_al_2015.h5
+title = depth = 1000000; colormap = PuRd; min_value = 0;
+        max_value = 80
```
joachimwolff, line 19: Is there a reason for the additional space?

lldelisle: It is to make it part of the title not a new parameter but I can put all in the same line if it is easier to understand.

pygenometracks/tracksClass.py
```diff
@@ -682,23 +698,21 @@ def plot(self, ax, chrom_region, region_start, region_end):
...
+        ax.set_xticks(ticks)
+
+        # Fix the limits which can be overwritten by set_xticks:
+        ax.set_xlim(start, end)
+
+        # Put the label
+        ax.set_xticklabels(labels)
 
-        ax.axis["x"].axis.set_ticklabels(labels)
         ax.axis['x'].axis.set_tick_params(which='minor', bottom='on')
```
joachimwolff, line 710: Not sure if this is our fault but this basically reads 'axis on axis on axis apply xyz'. Maybe that naming scheme can be changed?

lldelisle: In fact I don't know what this line is doing because I don't think we have minor ticks on the x-axis...

pygenometracks/tests/test_data/browser_tracks_hic_force_scale.ini
joachimwolff: I think it would be better to write it in quotation marks. 

```
title = 'depth = 500000; colormap = PuRd; min_value = 5;
        max_value = 70'
```

Currently it is not that obvious for users where the difference of this max_value is with e.g. the one in line 21. I fear if we leave it this way many will be confused.

lldelisle: If I put the quote or the `\n`:
image
I get them in the title:
image
The only way to go back to line is to use indentation: 
https://docs.python.org/3.4/library/configparser.html?highlight=configfile#supported-ini-file-structure
"Values can also span multiple lines, as long as they are indented deeper than the first line of the value."
But for clarity I will put all titles in only one line.

joachimwolff (approved):

PR #304, opened 2020-10-29
log.warn to log.warning

lldelisle:
To solve a deprecation warning.

### DIFF
--- a/pygenometracks/tracksClass.py
+++ b/pygenometracks/tracksClass.py
@@ -552,8 +552,10 @@ class PlotTracks(object):
                 file_field_name = key
                 # # THIS COULD BE REMOVED IN A NEXT 1.0 VERSION
                 if file_field_name == 'boundaries_file':
-                    log.warning("The boundaries_file is not used anymore"
-                                " please use another track with the"
+                    log.warning("Deprecation Warning: "
+                                "The boundaries_file is not used anymore."
+                                " It will be ignored."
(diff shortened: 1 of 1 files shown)


pygenometracks/tracksClass.py
```diff
@@ -552,9 +552,9 @@ def check_file_exists(track_dict, tracks_path, is_hic=False):
                 file_field_name = key
                 # # THIS COULD BE REMOVED IN A NEXT 1.0 VERSION
                 if file_field_name == 'boundaries_file':
-                    log.warn("The boundaries_file is not used anymore"
-                             " please use another track with the"
-                             " `overlay_previous` option.\n")
+                    log.warning("The boundaries_file is not used anymore"
```
joachimwolff, line 555: Also a Deprecated Warning?

lldelisle: In fact this PR is just changing all log.warn to log.warning because log.warn is deprecated since python 3.3 so to anticipate that once it would not work I decided to change all log.warn to log.warning.

lldelisle: This one is indeed a deprecation warning for PGT users (boundaries_files are not used anymore). You want me to put "Deprecation Warning" in the message?

joachimwolff: Yes, the latter one was my intend.

joachimwolff (approved):

PR #305, opened 2020-10-29
Support Matplotlib 3.3.2 and include python 3.8 in travis

lldelisle:
Implement tests for python 3.8
Redo all plots for matplotlib 3.3.2
Solve some warnings from matplotlib


pygenometracks/tests/test_non_classical_bed.py
```diff
@@ -178,7 +180,7 @@ def test_plot_tracks_bed_unusual_format():
            f"--outFileName {outfile.name}".split()
     pygenometracks.plotTracks.main(args)
     res = compare_images(expected_file,
-                         outfile.name, tolerance)
+                         outfile.name, 19)
```
joachimwolff, line 183: I think it is better use 'tolerance' and either add a specific number, increase tolerance globally  or provide a test file for python 3 (???). 

I thought we only support python 3, therefore I don't understand the comment to this test case (line 169, 170)

pygenometracks/tracks/HiCMatrixTrack.py
```diff
@@ -215,12 +216,10 @@ def set_properties_defaults(self):
                                                     shape=self.hic_ma.matrix.shape)
             self.hic_ma.matrix = self.hic_ma.matrix + main_diagonal
 
-        self.norm = None
-
         self.process_color('colormap', colormap_possible=True,
                            colormap_only=True, default_value_is_colormap=True)
 
-        self.cmap = cm.get_cmap(self.properties['colormap'])
+        self.cmap = copy.copy(cm.get_cmap(self.properties['colormap']))
```
joachimwolff, line 222: Please double check if you want to have shallow copy or a deep one: https://docs.python.org/3/library/copy.html

pygenometracks/tests/test_non_classical_bed.py
lldelisle: There is a small change in the font width between python 3.6, python 3.7 and python 3.8. Good idea to add a number to tolerance.

requirements.txt
```diff
@@ -1,5 +1,6 @@
 numpy >=1.16
-matplotlib ==3.1.1
+# matplotlib >=3.1.1,<=3.3.2 # For conda
+matplotlib ==3.3.2 # For the tests on travis
```
joachimwolff, line 3: Please use a patch or a sed command in the travis file, but keep the requirements.txt like it is in line 2.

joachimwolff: Or use a requirements-travis.txt file

lldelisle: I thought requirements.txt was only for travis...

joachimwolff: There are people like me installing (especially for development) the required dependencies to an environment using `conda install --file requirements.txt` and `python setup.py install` :)

lldelisle: Me too but then when you are developing you want the version as in travis, no? I don't mind to split it...

pygenometracks/tracks/HiCMatrixTrack.py
lldelisle: I followed the suggestion from matplotlib warning: 
```
MatplotlibDeprecationWarning: You are modifying the state of a globally registered colormap. In future versions, you will not be able to modify a registered colormap in-place. To remove this warning, you can make a copy of the colormap first. cmap = copy.copy(mpl.cm.get_cmap("bone_r"))
    self.cmap.set_bad('black')
```

joachimwolff (approved):

PR #308, opened 2020-11-03
Remove eval in process_color and check the operation before eval it

lldelisle:
I rewrote process_color to use regex instead of eval.
For operation, I do not see how to remove the eval but I checked that what was in the operation was only `allowed_words` and `allowed_signs`.
Please check that I did not forget any allowed sign or word...


lldelisle:
@bgruening  if you have time to give us your opinion on this... I did not realized that by putting eval at 2 different places (process_color) and (operation) I could allow code to be run on remote servers.
I think this solves the issue but tell me if you still see a flaw.

joachimwolff (approved):

PR #310, opened 2020-11-04
new parameters for scale_bar: scalebar_start_position scalebar_end_position

lldelisle:
To be able to place more easily a scalebar...


joachimwolff:
The name x_boundary1 is not intuitive for me. It only for the scalebar, why not `x_boundary_scalebar` or similar?

lldelisle:
Sure I am very bad at naming. But we can specify 2 x_boundary so `x_boundary_scalebar_1`? I don't want to use left and right because I want to have them symetric in case we use decreasing X axis.

lldelisle:
This is how it looks now:
master_scale_bar_startend_dec
master_scale_bar_startend

joachimwolff (approved):

PR #315, opened 2020-11-12
Release 3.6

lldelisle:


LeilyR (approved):

lldelisle:
# 3.6


## Enhancements:

- A new parameter for pyGenomeTracks allow to fix the width of the central part: ``--plotWidth``

- For scale bars, you can now define the position using 2 parameters among ``x_center``, ``size``, ``scalebar_start_position`` and ``scalebar_end_position``.


## Python dependency with conda:

- Using conda which is the way we recommand installation it is now possible to use Python 3.8.

## Relax version control on matplotlib:

- Any version of matplotlib from 3.1.1 to 3.3.2 is now accepted.

## Bugfixes:

- When the ``--fontSize`` parameter was used it happened that the colorbar labels were not corresponding to the colorbar ticks. This bug has been fixed (#297).

- The colormap ``pink`` was not working because it was both the name of a color and a colormap. This bug has been fixed (#295).

- When setting ``min_value``, ``max_value``, if the track was empty these values were ignored. This bug has been fixed.

- When using ``operation`` in bedgraph with a ``second_file``, BEDTools is now used to intersect the second file. ``use_middle`` is now taken into account on the second file.

- Different handling of string evaluation of color, colormap and operation.

joachimwolff (approved):

PR #317, opened 2020-11-19
cosmetic changes in attempt to make the code structure clearer

LeilyR:
cosmetic changes in attempt to make the code structure clearer


lldelisle:
Hi,
I am just wondering if for retrocompatibility issues, `bin/make_tracks_file` should not print a message that it is now called `make_tracks_config`.
Also I think in the parser we can change `--tracks` to `--tracksConfig`. `--tracks` will still work but in our examples we use `--tracksConfig` which is more clear.

PR #335, opened 2021-04-20
Setting output to required

joachimwolff:
If it is not set, the software crashes. Either we have it required or we provide a default file name


lldelisle:
Good catch. Should be targeted to develop branch no?

lldelisle (approved):

PR #351, opened 2021-07-16
Azure pipelines

joachimwolff:
I propose to move the testing to Microsoft Azure. The test cases pass on Azure, while on travis, they fail on macOS caused by a faulty installation. Also, the runtime on Azure is around three times faster.


lldelisle (approved):
Thank you very much

PR #354, opened 2021-08-11
Implement RNA-DNA interaction tracks

sansere:
Hi,

I would like to visualize bedpe files from iMARGI/RADICL etc. data in combination with other sequencing data from independent coordinates and chromosomes at once. For example DNA at chr1 and RNA at chr 7.
I attached an example 
DNA-RNA-interaction track


lldelisle:
Hi @sansere,
If you want that we implement a new feature, you need to write an issue. I opened one for you: #355.
A pull request is when you wrote the code and you want us to integrate it.

Cheers.

PR #359, opened 2021-08-19
Naive FastaTrack

lldelisle:
Here is an implementation of a FastaTrack thanks to @GDelevoye.
The whole fasta is stored in memory which can be an issue with large fasta files.


lldelisle:
Closed by #374

PR #360, opened 2021-08-19
FastaTrack with pyfaidx

lldelisle:
Very similar to #359 but uses pyfaidx.
Pro:
- Enable quicker access to a region

Cons:
- Require 'well formated fasta files': 
`pyfaidx.FastaIndexingError: Line length of fasta file is not consistent! Inconsistent line found in >rDNA_unit_8919x2_bp at line 3.`
- Add a new dependency


lldelisle:
I just realized that pyfaidx is already a dependency of cooler so of pgt...

lldelisle:
Closed by #374

PR #366, opened 2021-12-10
Hic matrix square

lldelisle:
I implemented a new track type which is like hicPlotMatrix.
I called it HiCMatrixSquareTrack. In ini file you need to put `file_type = hic_matrix_square`.
I am very bad at naming so if you want to change I have no objection.
I decided to create an abstract class HiCMatrixLikeTrack with 2 children: HiCMatrixTrack and HiCMatrixSquareTrack to avoid code duplications.
I think it would be good to implement loop_square and TAD_square to be able to overlap annotations on this.
I would have liked to be able to put a y axis in addition to the colorbar but this seems not easy before we implement real margin on the left.


joachimwolff (commented):
Why is the last matrix in the test data having curves?

lldelisle:
You mean that:
image

lldelisle:
I guess it is because too long pairs have been filtered out before making the cool file...

joachimwolff (approved):

lldelisle:
I don't understand why some linting are failing. In addition, they refer to a line which does not exists. @joachimwolff can you relaunch the CI?

PR #367, opened 2021-12-15
add new links_type (squares)

lldelisle:
This is mainly to overlay with hic_matrix_square


joachimwolff (commented):
Can you provide a plot with Hi-C + overlay of these squares?

lldelisle:
I already did it and it worked. See:
https://github.com/lldelisle/scriptsForBoltEtAl2022/blob/main/plots/outputs/fig6b.pdf
If both PR are approved I will make an example in the tests.

joachimwolff (approved):

PR #388, opened 2022-02-23
Deploy with azure

lldelisle:
This will deploy on pypi with my account when merged to master.


lldelisle:
I updated the markdown for pypi but this assumes we continue to tag release with the exact version number (no v in front of it).

joachimwolff (approved):

PR #390, opened 2022-02-23
fix append to potentially None

lldelisle:
This was not failing the actual tests but if someone wants to initiate tracks (through API) without regions some were failing.


joachimwolff (approved):

lldelisle:
Do you think I need to put a test for this or any how all this will change in 4.0

PR #391, opened 2022-02-24
Solve matrix inbetween

lldelisle:
Solve issues when plotting region which have no data but is not the end of the chr


joachimwolff (approved):

lldelisle:
I put it here in case we need it. Here is how I created the cool file:
```
echo -e "chrY\t91744698" > chry.size
echo -e "chrY\t31000000\tchrY\t31000000" > single_interaction.txt
cload pairs -c1 1 -p1 2 -c2 3 -p2 4 chry.size:100000 single_interaction.txt single_interaction_far_from_start.cool
```

PR #396, opened 2022-03-01
Mkdir

lldelisle:
Create the directory of the output file if it does not exists.

### DIFF
--- a/pygenometracks/plotTracks.py
+++ b/pygenometracks/plotTracks.py
@@ -142,6 +142,7 @@ type = vlines
 """
 
 import sys
+import os
 import argparse
 import warnings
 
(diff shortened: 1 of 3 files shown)


pygenometracks/plotTracks.py
```diff
@@ -275,6 +276,10 @@ def main(args=None):
                      track_label_width=args.trackLabelFraction,
                      plot_regions=regions, plot_width=args.plotWidth)
 
+    # Create dir if dir does not exists:
+    # Modified from https://stackoverflow.com/questions/12517451/automatically-creating-directories-with-file-output
+    os.makedirs(os.path.dirname(os.path.abspath(args.outFileName)), exist_ok=True)
```
joachimwolff, line 281: https://docs.python.org/3/library/os.html#os.makedirs

What does `confused` mean? `[makedirs()](https://docs.python.org/3/library/os.html#os.makedirs) will become confused if the path elements to create include [pardir](https://docs.python.org/3/library/os.html#os.pardir) (eg. “..” on UNIX systems).`

lldelisle: I will make a test. With an output which contains `../` or `./`. And come back to you.

lldelisle: It works. `os.path.abspath` resolves any `..` or `.`.

lldelisle: I don't know what 'confused' mean because I tried:
```
>>> os.makedirs("../toto")
>>> os.makedirs("pygenometracks/../toto")
>>> os.makedirs("pygenometracks/../toto/tralal/../t")
```
And all worked. Maybe the last one got an unexpected behaviour as it created both toto/tralal and toto/t.

joachimwolff (approved):

Issue #409, opened 2022-08-18
Error after installing pyGenomeTracks - ImportError H5Pget_fapl_direct


gene-drive:
Hi, I installed pyGenomeTracks using conda and mamba:

```
conda create -n pygenometracks -c bioconda -c conda-forge mamba python=3.7
conda activate pygenometracks
mamba install -c defaults -c bioconda -c conda-forge pygenometracks
```

Everything looked like it went well, however this is the error I'm seeing:
```
pyGenomeTracks --help
Traceback (most recent call last):
  File "/home/user_vm/anaconda3/envs/pygenometracks/bin/pyGenomeTracks", line 4, in <module>
    from pygenometracks.plotTracks import main
  File "/home/user_vm/anaconda3/envs/pygenometracks/lib/python3.7/site-packages/pygenometracks/plotTracks.py", line 149, in <module>
    from pygenometracks.tracksClass import PlotTracks
  File "/home/user_vm/anaconda3/envs/pygenometracks/lib/python3.7/site-packages/pygenometracks/tracksClass.py", line 17, in <module>
    from pygenometracks.tracks.GenomeTrack import GenomeTrack
  File "/home/user_vm/anaconda3/envs/pygenometracks/lib/python3.7/site-packages/pygenometracks/tracks/__init__.py", line 7, in <module>
    mod = __import__('.'.join([__name__, py]), fromlist=[py])
  File "/home/user_vm/anaconda3/envs/pygenometracks/lib/python3.7/site-packages/pygenometracks/tracks/HiCMatrixSquareTrack.py", line 3, in <module>
    from . HiCMatrixLikeTrack import HiCMatrixLikeTrack
  File "/home/user_vm/anaconda3/envs/pygenometracks/lib/python3.7/site-packages/pygenometracks/tracks/HiCMatrixLikeTrack.py", line 1, in <module>
    from hicmatrix import HiCMatrix
  File "/home/user_vm/anaconda3/envs/pygenometracks/lib/python3.7/site-packages/hicmatrix/HiCMatrix.py", line 14, in <module>
    import cooler
  File "/home/user_vm/anaconda3/envs/pygenometracks/lib/python3.7/site-packages/cooler/__init__.py", line 14, in <module>
    from .api import Cooler, annotate
  File "/home/user_vm/anaconda3/envs/pygenometracks/lib/python3.7/site-packages/cooler/api.py", line 11, in <module>
    import h5py
[...]

PR #410, opened 2022-08-22
solve issue #409 of install

lldelisle:
Solves #409 because in the default channel there is hdf5 with mpi which is incompatible with pyh5 nonmpi


lldelisle:
This PR goes to master as it is only documentation (and few CI fixes).

joachimwolff (approved):

PR #412, opened 2022-08-22
Fixes vhighlights features

lldelisle:
Following @joachimwolff advice, allow to put vhighlights in the background and remove the border.


pygenometracks/tracksClass.py
```diff
@@ -464,10 +467,21 @@ def parse_tracks(self, tracks_file, plot_regions=None):
                 else:
                     # We set the default value of vhighlight to 0.5
                     track_options['alpha'] = 0.5
+                if 'zorder' in all_keywords:
```
joachimwolff, line 470: Does that really remove the border? I had to apply some weird hacks to achieve this. As far as I remember I had to check the alpha value and set depending on it the alpha value for the border. I need to check my source code again.

pygenometracks/tracksClass.py
```diff
@@ -381,11 +381,14 @@ def plot_vhighlight(self, axis_list, chrom_region, start_region, end_region):
...
                     color = DEFAULT_VHIGHLIGHT_COLOR
+            # Process zorder:
+            if 'zorder' not in properties:
+                properties['zorder'] = -100
             for ax in axis_list:
                 for region in sorted(int_tree[chrom_region][start_region - 10000:end_region + 10000]):
                     ax.axvspan(region.begin, region.end,
-                               color=color, alpha=properties['alpha'])
-
+                               facecolor=color, alpha=properties['alpha'],
```
lldelisle, line 390: This is what removes the border

joachimwolff (approved):

PR #414, opened 2022-08-23
More options for vhighlight and vlines

lldelisle:
This is a big change in code structure.
types looks now like tracks which make the code more readable.
In addition, the vlines are now a single segment.
This is on top of #412


lldelisle:
I fixed the issue with matplotlib 3.1.1

joachimwolff (approved):

Issue #422, opened 2022-12-01
Links bug with use_middle


Jeff1995:
Hi developers! Thanks for developing this great tool. I have been using pyGenomeTracks to visualize links file with "use_middle" mode in a [package](https://github.com/gao-lab/GLUE) I developed. However, it occasionally raises error like this: gao-lab/GLUE#52.

I have looked into the error and spotted the reason.

When the linked regions [start1, end1] encompasses [start2, end2], there can be situations where:

* start1 < start2
* end1 >> end2

In which case we can have mid1 > mid2, producing a null interval in the following line:

https://github.com/deeptools/pyGenomeTracks/blob/2085e04a4bdea7690b52618d5dbe62410376992c/pygenometracks/tracks/LinksTrack.py#L502

Jeff1995:
I have made the above PR to fix this. Could you review it and see if that's the way to go?

lldelisle:
Hi,
Thank you very much, yes perfect.

PR #423, opened 2022-12-01
Fix links track with use_middle

Jeff1995:
Proposed bug fix for #422.


lldelisle:
Thank you very much for the bug fix, indeed, I did not think about this case.
What is missing to this PR is:
- target to develop instead of master and rebase or merge develop
- add a test so we never reintroduce this bug
- add a release text to announce this bug fix
Tell me if you are too busy to do it, I can do it as well.

Jeff1995:
Sure! I have rebased and targeted to develop, and modified a test file to include a situation that would have raised the error. Release text has also been added. Note that the modifed test file "test_wide.arcs" causes change to three test images. Could you check and see if that's appropriate?

lldelisle:
The test looks good thank you. I am just wondering if the output of master_arcs_use_middle.png should not be changed to show like we see on master_links_squares_overlay.png...

Jeff1995:
I guess this is just a corner case where the triangular and square visualizations look awkward, and should not have been used in the first place.

lldelisle:
Yes, I think you are right, we never thought that end2 could be smaller than end 1....
I guess the expected behaviour would be to go from start1 to max(end1, end2)

lldelisle:
What I was mentioning is what I draw in pink which is missing:
image
Also because we never thought about the case end2>end1

Jeff1995:
Thanks! Now I understand 😀. I agree with these expected behaviors. I'll see what I can do and get back to you tomorrow.

lldelisle:
Thank you very much. If you don't have time, I can do it.

Jeff1995:
I think I've got it to work.

Screen Shot 2022-12-02 at 14 22 06

lldelisle:
This is super great! Thank you so much.

docs/content/releases/3.7.1.rst
```diff
@@ -5,4 +5,4 @@ Bugfix:
 ^^^^^^^
 
 - Works with gffutils version 0.11.0 where ValueError became gffutils.exceptions.EmptyInputError for empty gtf.
-
+- Fixed a small bug in links track with "use_middle" enabled, where midpoints could have different ordering than start points.
```
lldelisle, line 8:
```suggestion
- Fixed a small bug in links track with "use_middle" enabled, where midpoints could have different ordering than start points (Thanks to @Jeff1995 ).
```

lldelisle (approved):

Issue #404, opened 2022-07-07
Support for mcool files with hic_matrix_square


mxw010:
**Welcome to the pyGenomeTracks GitHub repository! Before opening the issue please check
 that the following requirements are met :**

  - [ x ] Search whether this issue (or a similar issue) has been solved before
  using the search tab above. Link the previous issue if appropriate below.

  - [ x ] Paste your pyGenomeTracks version (`pyGenomeTracks --version`) and your python
  version (`python --version`) below.

```
pyGenomeTracks --version
pyGenomeTracks 3.7

python --version
Python 3.9.12
```

Pretty sure this is due to the lack of support for .mcool file for some reason, because when I have 

```
[hic matrix]
file = path-to-mcool.mcool::/resolutions/50000
```

in the .ini file I get 

```pyGenomeTracks --tracks test.ini --region chr2:222000000-223000000 -o tads_square.pdf
Traceback (most recent call last):
  File "~/.conda/envs/hicexplorer/lib/python3.9/site-packages/pygenometracks/tracksClass.py", line 677, in check_file_exists
    open(file_name_to_check, 'r').close()
FileNotFoundError: [Errno 2] No such file or directory: '~/path-to-file.mcool::/resolutions/50000'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "~/.conda/envs/hicexplorer/lib/python3.9/site-packages/pygenometracks/tracksClass.py", line 685, in check_file_exists
    open(name_with_tracks_path_to_check, 'r').close()
FileNotFoundError: [Errno 2] No such file or directory: '/~/path-to-file.mcool::/resolutions/50000'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "~/.conda/envs/hicexplorer/bin/pyGenomeTracks", line 11, in <module>
    main(args)
  File "~/.conda/envs/hicexplorer/lib/python3.9/site-packages/pygenometracks/plotTracks.py", line 274, in main
    trp = PlotTracks(args.tracks.name, args.width, fig_height=args.height,
  File "~/.conda/envs/hicexplorer/lib/python3.9/site-packages/pygenometracks/tracksClass.py", line 80, in __init__
    self.parse_tracks(tracks_file, plot_regions=plot_regions)
  File "~/.conda/envs/hicexplorer/lib/python3.9/site-packages/pygenometracks/tracksClass.py", line 604, in parse_tracks
    track_options = self.check_file_exists(track_options,
[...]

PR #424, opened 2022-12-06
Release 3.8

lldelisle:
Potential release text:

# 3.8

## Enhancement:

- A new ``type`` was introduced in 3.7 (announcement has been forgotten): a ``vhighlight`` to be able to hightlight a whole region across all tracks.
- Much more parameters for vhighlight and vlines
    - 'zorder' to set it on top or behind.
    - 'line_width', 'color', 'alpha', 'border_color' for vhighlight and 'line_style' for vlines
- You can now set multiple vlines
- For gtf, the output of ``make_tracks_file`` set ``prefered_name`` to ``gene_name`` instead of ``transcript_name``.

## Bugfix:

- Support mcool files for hic_matrix_square. This solves #404 
- Works with gffutils version 0.11.0 where ValueError became gffutils.exceptions.EmptyInputError for empty gtf.
- The installation instructions have been updated
- Fixed a small bug in links track with "use_middle" enabled, where midpoints could have different ordering than start points (Thanks to @Jeff1995 ).
- Fix a bug when plotting gtf in the middle of a large intron.
- Fix plot genes with ``orientation = inverted``
- Support Numpy 1.24, drop support below 1.20

## Relax version control on matplotlib:

- Any version of matplotlib from 3.1.1 to 3.6.2 is now accepted.


LeilyR (approved):
@lldelisle thanks!

lldelisle:
There is a new build for cooler but the CI does not use it... I will wait for Monday.

lldelisle:
It seems that mamba does not want to use the last build except if we specify `numpy >=1.16, <1.24`... Do you think we should write it into the dependencies?

lldelisle:
I need to update HiCMatrix before we merge it.

bgruening (approved):

PR #447, opened 2023-05-17
Add demo figure

lldelisle:
I think, this could go to master because it is just adding a new example in the documentation.


lldelisle:
I switched to develop as I need to incorporate the change in python_version to make tests passing...

bgruening:
Thanks a bunch!

PR #469, opened 2024-03-05
Fix lint isinstance

lldelisle:


bgruening:
Maybe we need to install `importlib_metadata`?

lldelisle:
Yes but I feel like the problem should not be solved in pyGenomeTracks but on HiCMatrix.

lldelisle:
I think we should drop support for python 3.7

bgruening:
Fine by me!

lldelisle:
I tried to avoid python 3.7 support drop but conda could not solve the environment so I gave up.

PR #473, opened 2024-05-03
Relax matplotlib versions compatible

lldelisle:
Everything worked smoothly except Arc which needed keyword parameters.


bgruening (approved):
Cool, that the tests still pass. I thought we are way more sensitive to matplotlib changes.

Thanks @lldelisle

PR #486, opened 2024-11-22
Update examaple: epilogos_track.ini

telatin:
Enable the categories file otherwise ignored


lldelisle:
@telatin thank you very much for finding this inconsistency. I updated the plot accordingly.

lldelisle:
Finally, looking at the docs: https://pygenometracks.readthedocs.io/en/latest/content/examples.html#examples-with-epilogos
I feel like it is more the title that should be changed to explain that it is a plot with default parameters.

PR #492, opened 2025-09-29
Remove anaconda usage

lldelisle:
This PR get rid of anaconda usage and use micromamba.
Additionally it pins a max version for bx-python in order to make tests pass.

I think could be merged to master as it only involves CI.


lldelisle:
@bgruening would you have time to review this one (only 2 files to fix the CI)?

lldelisle:
Thank you

PR #493, opened 2025-10-20
Add new track type: GwasTrack

AnttonLA:
Hi again,

I went back and re-visited the implementation of my proposed "GWAS track". This is a minimal implementation. I'll be very happy to build it up further, but I wanted to check that I was doing things right first.

## Description 
The track takes data from an IGV `.gwas` file (see format [here](https://igv.org/doc/desktop/#FileFormats/DataTracks/#gwas)), and draws a scatterplot where the x-axis is the genomic position and the y-axis is the negative log of the p-value (so the smaller the exponent, the higher it will appear).
This is the same as a localized Manhattan plot. It is useful when trying to visualize GWAS variants in relation to nearby genes, open chromatin, etc., and it's common to illustrate loci of interest in this way with tools like [LocusZoom](http://locuszoom.org/). The UCSC browser also has a built-in track with hits from the GWAS Catalog specifically to do lookups like this.

Example image:
new_track

## Implementation details
I tried to keep things small and match the style of the other tracks.
I've added a GwasTrack.py script, a corresponding test script, and test and example files.

The one thing I'm unsure about is the reading to the data from the file. I wanted to avoid additional dependencies, so created a  `ReadGwas` class. It is inspired on the 'readBed.py' file in the repo. Please let me know if I should use another approach instead, like keeping the file reading logic to `GwasTrack.py` or using a dependency like Pandas to keep the file-reading neat.

Thanks!


lldelisle:
Hi @AnttonLA ,
Thank you so much for this new track and the example files. I edited it to better fit with other tracks and add more customization, see #503 . I hope you like it.

AnttonLA:
Nice! Let me know if anything else is needed

